PR #49, opened 2023-10-17
Second attempt to fix flaky test

Ostrzyciel:
Not sure why these two specs fail, maybe because of the different execution context based on Pekko?


Ostrzyciel:
recheck

Ostrzyciel:
It does appear to be more stable. I got a strange failure on the first run with Java 11, but it is probably (???) unrelated.

PR #185, opened 2024-10-19
Jena & RDF4J: autodetect if parsed file is delimited or not

Ostrzyciel:
Continuation of #184. Jena RIOT and RDF4J Rio parsers will now first examine the first 3 bytes of the stream to determine if it's delimited or not. In the non-delimited case (previously unsupported), the file is assumed to contain exactly one stream frame.

The autodetection is fortunately unambiguous. After examining 3 bytes, we know for sure if this is one or the other.

On top of support for this functionality in Jena, I've added the same for RDF4J and added a bunch of integration tests.


Ostrzyciel:
The main docs should also be updated with this.

PR #187, opened 2024-10-21
Fix typo in docstring

Ostrzyciel:


Ostrzyciel:
Not sure why the CI is hanging like this. Maybe something with sbt or GH Actions?

Before investigating I think I'll just wait and see if it fixes itself... this works surprisingly often with github.

PR #193, opened 2024-10-22
Encoder: Maybe speed up stream row buffer

Ostrzyciel:
I thought that ArrayBuffer in Scala is smart enough to NOT allocate things if it still has existing capacity, but somehow in the profiler I'm reading a lot of allocations there. Let's try a custom implementation of the row buffer – as simple as possible.

If this works, I will finish the docstrings. If not, I will revert it.


Ostrzyciel:
+ We should also add tests for the FastBuffer class, to be 100% sure it works correctly with expanding capacity...

PR #206, opened 2024-11-03
Experiment: speed up prefixed IRI encoding

Ostrzyciel:
This is an experimental rewrite of prefixed IRI encoding that aims to reduce the number of hashmap calls by merging the complete IRI map with the name lookup table. The name lookup table is now keyed by IRIs (not names), which is based on the assumption that names are often associated with only one prefix. Of course, this assumption can be broken. The code has a path to detect this if two such names occur one after another and deduplicates them. However, if the two IRIs are separated by another IRI, a duplicated name entry will be added.

It's an open question as to what would be exactly the performance and the representation size impact of this change. It's certain that the compression factors for datasets like `assist-iot-weather` will be hurt, but I'm not sure by how much. As usual, if it turns out to not be worth it, I will revert the change.

### DIFF
--- a/core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
+++ b/core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
@@ -1,6 +1,9 @@
 package eu.ostrzyciel.jelly.core;
 
+import jdk.dynalink.linker.support.Lookup;
+
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 
 /**
  * A lookup table for NodeEncoder, used for indexing datatypes, IRI prefixes, and IRI names.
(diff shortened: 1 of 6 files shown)


core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
```diff
@@ -1,6 +1,9 @@
 package eu.ostrzyciel.jelly.core;
 
+import jdk.dynalink.linker.support.Lookup;
+
 import java.util.HashMap;
+import java.util.LinkedHashMap;
```
Ostrzyciel, line 7: Cleanup imports.

integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/CrossStreamingSpec.scala
```diff
@@ -104,8 +104,9 @@ class CrossStreamingSpec extends AnyWordSpec, Matchers, ScalaFutures, JenaTest:
               val is = new FileInputStream(sourceFile)
               val os = new ByteArrayOutputStream()
               var encSize = 0
+              var frames = new mutable.ArrayBuffer[eu.ostrzyciel.jelly.core.proto.v1.RdfStreamFrame]()
               encFlow.tripleSource(is, limiter, jOpt)
-                .wireTap(f => encSize += f.serializedSize)
+                .wireTap(f => { encSize += f.serializedSize ; frames += f })
                 .toMat(decFlow.tripleSink(os))(Keep.right)
```
Ostrzyciel, line 110: Remove debug code.

PR #208, opened 2024-11-05
Minor optimizations and cleanup in decoding/encoding

Ostrzyciel:
Cleaned up old comments and unnecessary instructions.

Applied branchless code in a few places of the decoder. This should have a mildly positive performance impact (especially on more modern CPUs), but don't count on it being huge.

### DIFF
--- a/core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
+++ b/core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
@@ -58,10 +58,10 @@ final class EncoderLookup {
     // This will monotonically increase until it reaches the maximum size.
     private int used;
     // The last id that was set in the table.
-    private int lastSetId;
+    private int lastSetId = -1000;
     // Names of the entries. Entry 0 is always null.
     private final String[] names;
-    // Whether to use serials for the entries.
+    // Whether to maintain serial numbers for the entries.
     private final boolean useSerials;
 
     private final LookupEntry entryForReturns = new LookupEntry(0, 0, true);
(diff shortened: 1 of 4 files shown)


core/src/main/java/eu/ostrzyciel/jelly/core/EncoderLookup.java
```diff
@@ -58,10 +58,10 @@ public LookupEntry(int getId, int setId, boolean newEntry) {
     // This will monotonically increase until it reaches the maximum size.
     private int used;
     // The last id that was set in the table.
-    private int lastSetId;
+    private int lastSetId = -1000;
```
bswck, line 61: Why -1000? To avoid [the <-128, 127> integer cache](https://docs.oracle.com/en/java/javase/23/docs/api/java.base/java/lang/Integer.html?is-external=true#valueOf(int))? @Ostrzyciel

bswck: I'm seeing that this was always a 1-based lookup https://github.com/Jelly-RDF/jelly-jvm/blob/5f75b05db86805c540c67a19f9461fce9c57ac98/core/src/main/scala/eu/ostrzyciel/jelly/core/EncoderLookup.scala#L10-L11

Ostrzyciel: Don't read that much into this code. :) it's not going to help you with Python, it's way more likely to be confusing.

The integer cache is irrelevant here. This is not what this will compile to.

-1000 because I like it. It's not important for the algorithm, it just must be invalid.

bswck: > The integer cache is irrelevant here. This is not what this will compile to.

Yeah, I initially looked at `Integer`, which is irrelevant to `int`.

bswck: Thanks!

PR #223, opened 2024-11-14
Jena CLI: allow changing writer settings with presets

Ostrzyciel:
Previously it was not really possible to change the Jelly writer options when using the Jena CLI utilities. There is a CLI option to modify the Jena context, but our context would have to be set to a Java object... which I honestly hope is not possible, because that would be RCE or something.

So, to get around this, this commit introduces "presets" based on the default setting presets available in JellyOptions. The user can simply choose one of these by setting a symbol to the name of the preset.

I've added a test that checks if this works across all writers.

I've also added a doc page documenting this, because honestly, it's a neat and useful feature.


Ostrzyciel:
I'll review and merge it tomorrow. Sleep.

Issue #244, opened 2024-12-21
Implement namespace declarations for the Pekko Streams API


Ostrzyciel:
Follow-up of #227 

In that task I did not implement an API for this in the `stream` module – the namespace declarations are currently entirely ignored.

I've looked initially into how this could be done, but for now I don't have anything satisfactory. Certainly I wouldn't want the current API to be broken by the changes... and I wouldn't want to introduce a bazillion new method variants. Maybe we could have some kind of flow builders, expanding on the idea already used in `DecoderFlow`?

In any case, I do not have any use case for this yet (the nanopublication one doesn't use Pekko anyway), so this is a low priority task for now.

Ostrzyciel:
Resolved in #256 ... mostly. The rest will be done in #257

PR #256, opened 2025-01-12
New API for Pekko Streams encoders

Ostrzyciel:
Issue: #244 + more

This is a complete rewrite of the stream encoding API in the `jelly-stream` module. The previous methods in `EncoderFlow` and `EncoderSource` were replaced with flexible builder APIs that give even more flexibility. Notably, now it is possible to include namespace declarations in the streams.

The previous APIs were deprecated and will be removed in 3.0.0.

After this change unfortunately in some cases client code will get a bit longer (esp. if you were using `EncoderSource`), but I don't see any sensible way to avoid that. The previous setup was impossible to extend, so when we got a new feature in Jelly (namespace declarations), we could not easily add it to the streaming API. This should not be a problem now, as it would be simply a matter of adding a method or two.

I've also done a few other minor changes, more or less overdue.

- Added `NamespaceDeclaration` utility case class, to avoid relying on tuples when passing namespace decls. in streams.
- Properly deprecated the old ProtoEncoder factory methods.
- Switched the streaming API to use the new external management mode of the row buffer, for possibly better performance.

Of course, the docs and examples were updated to the new API.

This PR does not include any changes to `DecoderFlow`... it would be too much in one go. I'm postponing this for later. The need there is definitely not as pressing as here.


core/src/main/scala/eu/ostrzyciel/jelly/core/IterableAdapter.scala
```diff
@@ -1,5 +1,6 @@
 package eu.ostrzyciel.jelly.core
 
+import scala.annotation.targetName
```
Ostrzyciel, line 3: remove

core/src/main/scala/eu/ostrzyciel/jelly/core/NamespaceDeclaration.scala
```diff
@@ -0,0 +1,3 @@
+package eu.ostrzyciel.jelly.core
+
+final case class NamespaceDeclaration(prefix: String, iri: String)
```
Ostrzyciel, line 3: add scaladoc

PR #269, opened 2025-02-05
Try to add Graal EA tests

Ostrzyciel:
Including enabled support for JEP 450... let's see if this blows up.


Ostrzyciel:
meh, it's not working. It's not important enough to keep digging, so let's stop here.

PR #278, opened 2025-02-17
core: completely refactor ProtoEncoder

Ostrzyciel:
`ProtoEncoder` used to be "the ugly" part of the code where the library implementations were supposed to extend the class, implement some protected methods, and then call some protected final ones... and so on.

Not only is this messy ("one big superclass") but it also makes it very hard to extend the code for something like Jelly-Patch: https://github.com/Jelly-RDF/jelly-protobuf/issues/11

This PR fixes this:

- Introduce the `ProtoEncoderConverter` trait, analogous to the one we have for decoders. Instances of this trait are passed to the actual encoder.
- Introduce a public interface for `NodeEncoder` to be used by the aforementioned converters. This removes one layer of indirection and makes it possible to decouple node encoding from the `ProtoEncoder` itself.
- Moved the actual implementation of `ProtoEncoder` to a private class and the `core.internal` package, along with a few other helper classes.

This will NOT change the external public interface for Jelly encoding, but it does change the interface used by RDF library implementations. I think we can make an exception from the semver rules here, as I'm not aware of any such external implementations, or intents for such implementations. So, this should go into a MINOR release.


core/src/main/scala/eu/ostrzyciel/jelly/core/ConverterFactory.scala
```diff
@@ -1,6 +1,7 @@
 package eu.ostrzyciel.jelly.core
 
-import ProtoDecoderImpl.*
+import eu.ostrzyciel.jelly.core.internal.ProtoDecoderImpl.*
+import eu.ostrzyciel.jelly.core.internal.ProtoEncoderImpl
```
Ostrzyciel, line 4: Are these imports needed?

core/src/main/scala/eu/ostrzyciel/jelly/core/ProtoDecoder.scala
```diff
@@ -1,5 +1,6 @@
 package eu.ostrzyciel.jelly.core
 
+import eu.ostrzyciel.jelly.core.internal.ProtoDecoderImpl
```
Ostrzyciel, line 3: Not needed. Replace with a FQN in the scaladoc.

core/src/main/scala/eu/ostrzyciel/jelly/core/internal/ProtoEncoderBase.scala
```diff
@@ -0,0 +1,46 @@
...
+      `object` = nodeToProtoWrapped(converter.getTstO(triple), lastObject),
+    )
+
+  protected final def quadToProto(quad: TQuad): RdfQuad =
+    RdfQuad(
+      subject = nodeToProtoWrapped(converter.getQstS(quad), lastSubject),
+      predicate = nodeToProtoWrapped(converter.getQstP(quad), lastPredicate),
+      `object` = nodeToProtoWrapped(converter.getQstO(quad), lastObject),
+      graph = graphNodeToProtoWrapped(converter.getQstG(quad)),
+    )
```
Ostrzyciel, line 46: missing newline

core/src/test/scala/eu/ostrzyciel/jelly/core/internal/TranscoderLookupSpec.scala
```diff
@@ -1,5 +1,6 @@
-package eu.ostrzyciel.jelly.core
+package eu.ostrzyciel.jelly.core.internal
 
+import eu.ostrzyciel.jelly.core.internal.TranscoderLookup
```
Ostrzyciel, line 3: not needed

Ostrzyciel (commented):
Also: check if the documentation for lib implementers is up to date.

Karolina-Bogacka (approved):
well done. Truly a complete refactor

PR #282, opened 2025-03-01
Add ProtoDecoder.ingestRowFlat

Ostrzyciel:
This method avoids the overhead of allocating Some[Statement] during parsing. The old .ingestRow method is kept both for back-compat and just because it's still useful.

You may want to use ingestRowFlat if you are working with Java code or want to squeeze out an extra bit of performance.


Karolina-Bogacka (approved):
Ok, I think the purpose of these changes is clear

PR #292, opened 2025-03-07
Implement the RdfStreamFrame metadata field

Ostrzyciel:
Issue: https://github.com/Jelly-RDF/jelly-protobuf/issues/32

This is only a low-level API implementation, because currently there is no need to implement this in RDF4J, Jena, or the stream APIs. There is full support in the transcoding API, though.

To keep the protoVersion constants tidy, I've changed their naming scheme and deprecated one of the old ones.


core/src/main/scala/eu/ostrzyciel/jelly/core/ProtoEncoder.scala
```diff
@@ -24,6 +24,7 @@ object ProtoEncoder:
     options: RdfStreamOptions,
     enableNamespaceDeclarations: Boolean = false,
     maybeRowBuffer: Option[mutable.Buffer[RdfStreamRow]] = None,
+    enableStreamFrameMetadata: Boolean = false,
```
Ostrzyciel, line 27: Remove, not used

Karolina-Bogacka (approved):

Issue #294, opened 2025-03-10
Binding for titanium-rdf-api


vorburger:
@Ostrzyciel Thank You for chiming in over on https://github.com/eclipse-rdf4j/rdf4j/issues/5268!

Looking at this project, and noticing it already has both RDF4j & Apache Jena bindings, I just wanted to give you a shout-out here to https://github.com/filip26/titanium-rdf-api from https://github.com/apache/jena/discussions/2216 by @filip26.

Perhaps it would make sense to add a "binding" from/to Jelly to that API? Then  one could e.g. write a converter with https://github.com/filip26/iridium-cbor-ld. (Perhaps I'll do this in https://docs.enola.dev/use/rosetta/ one fine day.)

Ostrzyciel:
Hi! Thanks for the suggestion. Yeah, I think it would make sense.

You can currently do Jelly<->Titanium conversion by going through Jena, but admittedly that adds an additional conversion layer.

Do I understand correctly that your use case relies on titanium-rdf-api to represent RDF data in-between different RDF formats? This is interesting, I haven't previously heard of this library.

filip26:
Hi @Ostrzyciel 
You can find more details about the motivation and evolution of the RdfQuadConsumer interface [Extract RDF Interfaces to avoid conversions](https://github.com/apache/jena/discussions/2216). 

The RDF types implementation is intentionally missing, allowing you to use any implementation you prefer. Since this was formerly part of Titanium JSON-LD (which has its own RDF types implementation), that part will eventually be extracted into a separate artifact (e.g. `titanium-rdf-types`) that you can choose to combine with the consumer - or not.

Please feel free to raise any issues, contribute, or reach out. Thank you!

Ostrzyciel:
Tech note: see Jena implementation by afs for reference: https://github.com/apache/jena/pull/3061

Ostrzyciel:
@filip26 @vorburger  I started prototyping something, but I got confused over what would be expect from such a binding.

Jelly-JVM is by itself not an RDF library. It does not have the primitive types for RDF and by itself it does not have a serialization subsystem like RIOT or Rio. It's designed around generics, so that it can be easily integrated with libraries like Jena and RDF4J. If you look at the code of `jelly-jena` and `jelly-rdf4j`, it simply instantiates Jelly's core abstract classes with Jena's/RDF4J's types, and then integrates it with their I/O APIs.

[...]

PR #296, opened 2025-03-12
Revert "Add a term-level interface to ProtoEncoder (#295)"

Ostrzyciel:
This reverts commit 702fd63d3c76793e394528df775b21d1cc205cc6.

I was wrong, this is not needed to implement #294. I forgot that triple objects in this library are represented by a few strings, not a single string. So, we still need to allocate something to represent a node, and we can't skip the overhead.

Instead, I'd suggest to use the Titanium classes for RDF primitives. We have to use something, and I don't think it makes sense to introduce a new set of classes just for this integration.


Ostrzyciel:
Converting to draft. Not sure, maybe will be needed.

PR #297, opened 2025-03-13
Implement binding for titanium-rdf-api

Ostrzyciel:
First part of implementing #294 

Remaining work to do: documentation, README


filip26:
Great job! 

From a marketing perspective, it might be beneficial to name it `QuadConsumerReader` and `QuadConsumerWriter` (or something generic like that) to emphasize that it depends on a simple interface and is fully compatible with any library that supports the interface right out of the box.

Maybe, this could also serve as an easy way to bootstrap new users to Jelly, allowing them to explore and play with it without needing to deep dive into Jelly internals right away. Just thoughts.

Ostrzyciel:
@filip26 Thanks! 

Not sure about the naming, for me it'd be confusing. The N-Quads library calls its classes `NQuadsReader`. There is also already a `JellyWriter` implemented for Jena, so I wouldn't want to add to the confusion with that.

Good point about lowering the barrier to entry. I understand that this is a slight nudge about the general obtuseness of the Jelly website :laughing: yes, it could use some work. I initially wrote it to convince scientists, but that was a dead-end. It's time to adjust it.

.github/workflows/scala.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       shell: bash
       run: sbt -v +test
 
+      # TODO: move this test to a separate job, run only on Java 17
```
Karolina-Bogacka, line 45: why only on Java 17? out of curiosity

integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/IoSerDesSpec.scala
```diff
@@ -87,6 +87,11 @@ class IoSerDesSpec extends AnyWordSpec, Matchers, ScalaFutures, JenaTest:
     options.maxDatatypeTableSize should be (expOpt.maxDatatypeTableSize)
     options.version should be (Constants.protoVersion_1_0_x)
 
+  private def checkImplOptSupport(impl: NativeSerDes[?, ?], opt: Option[RdfStreamOptions]) =
```
Karolina-Bogacka, line 90: I would maybe like a tiny bit of a docstring for this - the conditions you check are self evident, but each of them is in the middle of a longer line of code, so you have to read in to check which conditions the function checks exactly

integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/IoSerDesSpec.scala
```diff
@@ -153,7 +178,8 @@ class IoSerDesSpec extends AnyWordSpec, Matchers, ScalaFutures, JenaTest:
             data.size should be > 0
             // In case we are leaving the default settings, RDF4J has no way of knowing if the data is
             // triples or quads, so it's going to default to quads.
-            val mayBeQuads = ser.name == "RDF4J" && preset.isEmpty
+            // Titanium just always does quads.
+            val mayBeQuads = ser.name == "RDF4J" && preset.isEmpty || ser.name == "Titanium"
```
Karolina-Bogacka, line 182: this IS nitpicking, but why not maybeQuads instead of mayBeQuads? should be easier to type and means roughly the same thing

titanium-rdf-api/src/main/java/eu/ostrzyciel/jelly/convert/titanium/TitaniumJellyEncoder.java
```diff
@@ -0,0 +1,56 @@
...
+     * Returns the number of rows currently in the encoded row buffer.
+     * @return int
+     */
+    int getRowCount();
+
+    /**
+     * Returns the rows in the encoded row buffer as a Scala collection and clears the buffer.
+     * @return scala.collection.immutable.Seq<RdfStreamRow>
+     */
+    scala.collection.immutable.Seq<RdfStreamRow> getRowsScala();
```
Karolina-Bogacka, line 43: are we sure we want getRowsScala and getRowsJava instead of a more general getRows(format) that could be then extended to include idk, Kotlin if needed?

titanium-rdf-api/src/main/scala/eu/ostrzyciel/jelly/convert/titanium/TitaniumJellyEncoderImpl.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  final override def getRowsScala: Seq[RdfStreamRow] =
+    val list = buffer.toList
+    buffer.clear()
+    list
+
+  final override def getRowsJava: java.lang.Iterable[RdfStreamRow] =
+    val list = buffer.toList
+    buffer.clear()
+    list.asJava
```
Karolina-Bogacka, line 63: ditto with my comment above, this two implemented methods differ only in the last line

titanium-rdf-api/src/test/scala/eu/ostrzyciel/jelly/convert/titanium/TitaniumJellyEncoderSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+  "TitaniumJellyEncoder" should {
+    "be created with default options" in {
+      val encoder = TitaniumJellyEncoder.factory()
+      encoder.getOptions should be (
+        JellyOptions.smallStrict
+          .withPhysicalType(PhysicalStreamType.QUADS)
+          .withLogicalType(LogicalStreamType.FLAT_QUADS)
+          .withVersion(CoreConstants.protoVersion_1_0_x)
+      )
+      encoder.getRowCount should be (0)
```
Karolina-Bogacka, line 24: may be good to add some test cases when the rowcount is larger than 0

Karolina-Bogacka (approved):
Minor comment, broadly looks good, happy to see integration tests :)

integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/IoSerDesSpec.scala
Ostrzyciel: OK, done.

titanium-rdf-api/src/main/java/eu/ostrzyciel/jelly/convert/titanium/TitaniumJellyEncoder.java
Ostrzyciel: No, we will never support Kotlin, at least not in this way.

The reason I made it like this is because the underlying data structure is from the Scala standard library, and to use it in Java you need a conversion. 

A general getRows(language) would be a massive headache :D let's not go there.

titanium-rdf-api/src/test/scala/eu/ostrzyciel/jelly/convert/titanium/TitaniumJellyEncoderSpec.scala
Ostrzyciel: OK, done.

integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/IoSerDesSpec.scala
Ostrzyciel: No, it's not "maybe quads", it's "may be quads" :D but that's old code anyway. This test will need a refactor anyway when we introduce test kits.

Ostrzyciel:
@filip26 I will stick with what we have now :) thank you for the help, though.

The receiver/emitter naming doesn't exactly fit what the reader/writer will do. They work just like any other RDF serializer. You stuff quads into the Writer, and it puts bytes in an OutputStream. You stuff InputStream bytes into the reader and get quads.

To do "real" RDF streams, like with Kafka, you'd need some heavier armament, like we have in the `jelly-stream` module: https://jelly-rdf.github.io/jelly-jvm/dev/user/reactive/
But that's a completely different API, and it won't work with Titanium. It requires that triples and quads are represented as objects.

I'll open a PR with docs in a moment, I hope it will clarify a few things.

PR #303, opened 2025-03-14
Never mark Rio outputs as using generalized RDF

Ostrzyciel:
RDF4J Rio simply does not have an API capable of producing generalized statements. It was my mistake including this option in the Rio integration: #222

However, it's still possible to serialize generalized statements using the new, low-level term-based API: #295


Karolina-Bogacka (approved):
Looks OK to me

PR #305, opened 2025-03-14
Add integration tests for generalized RDF

Ostrzyciel:
I've realized that there are no integration test cases for this, and really, the RDF library support for generalized RDF is patchy at best. These few tests at least check that it works on a basic level.

### DIFF
--- a/integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/GeneralizedRdfSpec.scala
+++ b/integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/GeneralizedRdfSpec.scala
@@ -12,7 +12,7 @@ import scala.reflect.ClassTag
 /**
  * Separate tests for handling generalized RDF.
  *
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/scala/eu/ostrzyciel/jelly/integration_tests/io/GeneralizedRdfSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+import scala.reflect.ClassTag
+
+/**
+ * Separate tests for handling generalized RDF.
+ *
+ * Normally this would be included with the IoSerDesSpec, but RDF libraries have very patch support
```
Karolina-Bogacka, line 15: spelling mistake, "have very patchy support" here

Karolina-Bogacka (approved):
I don't have much more to add, except I like this approach to these kinds of tests, roundTripTests worked out to be quite reusable here.

PR #307, opened 2025-03-16
Improve exception handling in core

Ostrzyciel:
This tidies up exception handling in decoders and encoders, reducing the chance that some NPE, array-out-of-bounds, class-cast, or whatever exception slips out. There would be nothing inherently *wrong* with that, as it's all unchecked exceptions anyway, but such exceptions are not very informative to users. By catching and re-throwing them, we can make it clearer why a failure has occurred.


Karolina-Bogacka (approved):
Generally looks good, the only feedback I can give is that I don't see any tests for the transcoding error and it might be worthwhile to add them

Ostrzyciel:
> Generally looks good, the only feedback I can give is that I don't see any tests for the transcoding error and it might be worthwhile to add them

There is no behavior change in that part, so the existing tests are fine.

PR #312, opened 2025-03-19
Docs: Improve clarity of index.md

Ostrzyciel:
Make the library-related links point to relevant user guides. Add more links to the guides in the module list.


bswck (approved):
LGTM.

<details>

<summary>On a side note (offtopic)</summary>

I think it would be cool if we had automatic preview builds attached to PR descriptions, so we don't need to `gh checkout <pr-number> && uvx mkdocs --with-requirements requirements.txt mkdocs serve` everytime to review the docs :p Do you like the idea, should I file an issue to track? It's not very hard to set up, there are some bots/actions that do that.

</details>

Ostrzyciel:
> I think it would be cool if we had automatic preview builds attached to PR descriptions, so we don't need to `gh checkout <pr-number> && uvx mkdocs --with-requirements requirements.txt mkdocs serve` everytime to review the docs :p Do you like the idea, should I file an issue to track? It's not very hard to set up, there are some bots/actions that do that.

Sure, why not. Go ahead.

PR #319, opened 2025-03-31
Unpin Python version in docs workflow

bswck:
Follows up #318. Forgot to check that before.
We can remove that line now, as the action understands the pin from `.python-version`.


Ostrzyciel (approved):

bswck:
Uhhh, no. I just checked the workflow report... It doesn't see the `.python-version` in path, because it's running from the repo root. But we can set the correct path in the workflow.

PR #321, opened 2025-04-02
Rio&RIOT: add option to write non-delimited output

Ostrzyciel:
Needed for: https://github.com/Jelly-RDF/cli/issues/55

This adds an option in the Rio and RIOT writers to disable frame delimiters. This is useful in some scenarios (if, e.g., the output stream is pointed to some internal buffer), or for development.

This should not be used by default for writing files to disk, as the entire frame (entire stream, in this case) would have to be kept in memory before being serialized. This is obviously not great.


Karolina-Bogacka (approved):
Looks useful and I see you have tests for Jena as well as RDF4J so looks good for me

PR #324, opened 2025-04-08
Make AnyStatementDecoder ignore duplicate options

Ostrzyciel:
This was the only decoder that threw an error when encountering duplicate stream options, which was pretty inconsistent. This was technically OK with the spec ("Implementations SHOULD NOT throw an error if the stream options header is present more than once in the stream."), but was not the RECOMMENDED behavior.

Encountered while working on: https://github.com/Jelly-RDF/cli/issues/66


Karolina-Bogacka (approved):
Seems small and simple enough, logic makes sense

PR #327, opened 2025-04-09
Match spec-defined behavior with small table sizes

Ostrzyciel:
Follow-up of https://github.com/Jelly-RDF/jelly-rdf.github.io/pull/35

Make the smallest allowed name table size be 8.

Allow 0-sized datatype tables, if no datatype literals are used.

Update tests to make sure this behavior is respected.


Karolina-Bogacka (approved):
Accept - the changes look like what's on the tin

PR #329, opened 2025-04-11
Enable dependabot

bswck:
For keeping actions & the jelly-protobuf submodule up-to-date semi-automatically


bswck:
Had to check if `directory` for actions were correct, but yeah, [they are](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot#enabling-dependabot-version-updates-for-actions).

Ostrzyciel:
Thanks!

Issue #288, opened 2025-03-05
Implement Jelly-Patch


Ostrzyciel:
Off-shoot of https://github.com/Jelly-RDF/jelly-protobuf/issues/11 to keep track of the implementation-specific issues.

- [x] Allow blank nodes as graph names in NS declarations (possibly also literals, quoted triples)
- [x] jelly-core-patch with low-level tests #333
- [x] jelly-jena-patch: converters #336
- [x] jelly-jena-patch: high-level APIs (reader, writer) #336
- [ ] Low-level behavior tests for jelly-jena-patch
- [ ] Jena integration with Fuseki – will require API extensions in Jena! #334
- [x] jelly-rdf4j-patch: converters #336
- [ ] jelly-rdf4j-patch: high-level APIs (reader, writer)
- [ ] Low-level behavior tests for jelly-rdf4j-patch
- [ ] Integration tests:
  - [x] Cross-streaming #336
  - [ ] Non-delimited test cases
  - [ ] Generalized RDF test cases
- [ ] Document it on the website
- [ ] Make some example code

Ostrzyciel:
I've just noticed that in the proto we only allow IRIs to be graph names in namespace declarations, because I forgot blank nodes are also valid as graph names. We should add that, but while we are at it, we can also add support for generalized RDF, to be consistent with the rest of Jelly. I added a point to do that at the top of the todo.

Ostrzyciel:
> I've just noticed that in the proto we only allow IRIs to be graph names in namespace declarations, because I forgot blank nodes are also valid as graph names. We should add that, but while we are at it, we can also add support for generalized RDF, to be consistent with the rest of Jelly. I added a point to do that at the top of the todo.

This was resolved in #361

PR #333, opened 2025-04-11
Experimental Jelly-Patch implementation

Ostrzyciel:
Issue: #288 

**This is just an experimental implementation – it's not a part of the public stable API!**

This is the core module for Jelly-Patch only, with no integrations for Jena or RDF4J. The encoder/decoder API is pretty similar to that in `jelly-core`, but with a few improvements to make it slightly cleaner.

There are unit tests for the core code. Integration tests will come later.

There are a few changes in `jelly-core` included here, so that we can reuse the code for Jelly-Patch. The changes are to the internal API only, so users should not notice any change.


core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoder.scala
```diff
@@ -0,0 +1,41 @@
...
+   * RdfPatchOptions for this decoder.
+   * @return options
+   */
+  def getPatchOpt: Option[RdfPatchOptions]
+
+  /**
+   * Ingest a row into the decoder.
+   *
+   * If the stream has type PUNCTUATED, and the row contains a punctuation mark, the decoder will
+   * call the punctuation() method on the handler.
```
Karolina-Bogacka, line 27: don't really get what you mean here. This is a trait that only specifies how Patch Decoders should look like (so any decoders implementing it will have to provide an ingestRow implementation. So the decoder SHOULD call the punctuation() method on the handler, but won't necessarily.

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/internal/PatchDecoderImpl.scala
```diff
@@ -0,0 +1,188 @@
...
+  private final def setPatchOpt(opt: RdfPatchOptions): Unit =
+    if streamOpt.isEmpty then
+      streamOpt = Some(opt)
+      isFrameStreamType = opt.streamType.isFrame
+      isPunctuatedStreamType = opt.streamType.isPunctuated
+
+  override final def ingestFrame(frame: RdfPatchFrame): Unit =
+    for row <- frame.rows do
+      ingestRow(row)
+    if isFrameStreamType then
```
Karolina-Bogacka, line 44: yeah, so in this case this checks out and handler.punctuation() is called, but in the trait I would phrase it differently, more in terms of shoulds / do sth similar to the run/doRun pattern to ensure it is always called if needed.

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/internal/PatchDecoderImpl.scala
```diff
@@ -0,0 +1,188 @@
...
+        convertTermWrapped(statement.`object`, lastObject),
+        convertGraphTermWrapped(statement.graph),
+      )
+
+  private[core] final class AnyStatementDecoder[TNode, TDatatype : ClassTag](
+    converter: ProtoDecoderConverter[TNode, TDatatype, ?, ?],
+    handler: AnyPatchHandler[TNode],
+    supportedOptions: RdfPatchOptions,
+  ) extends PatchDecoderImpl[TNode, TDatatype](converter, handler, supportedOptions):
+    private var inner: PatchDecoderImpl[TNode, TDatatype] = uninitialized
```
Karolina-Bogacka, line 151: not sure why this class needs an inner instance of PatchDecoderImpl when it already extends the same class and it is not a pattern I see in other decoders

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/internal/PatchDecoderImpl.scala
```diff
@@ -0,0 +1,188 @@
...
+          // both know how to handle punctuations.
+          handleOptions(opt)
+          createInnerDecoder(opt)
+          inner.ingestRow(row)
+      else
+        inner.ingestRow(row)
+
+    // Should not be called.
+    // We made these methods abstract to avoid the overhead of virtual function calls.
+    override def handleStatementAdd(statement: RdfQuad): Unit = ()
```
Karolina-Bogacka, line 173: I guess that it's because we move ingestRow logic to inner but why this specific pattern here

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
```diff
@@ -0,0 +1,354 @@
...
+@experimental
+class PatchDecoderSpec extends AnyWordSpec, Matchers:
+  val decoders: Seq[(
+    String,
+    (AnyPatchHandler[Mrl.Node], Option[RdfPatchOptions]) => PatchDecoder,
+    PatchStatementType
+  )] = Seq(
+    ("TriplesDecoder", MockPatchConverterFactory.triplesDecoder, PatchStatementType.TRIPLES),
+    ("QuadsDecoder", MockPatchConverterFactory.quadsDecoder, PatchStatementType.QUADS),
+    ("AnyDecoder", MockPatchConverterFactory.anyStatementDecoder, PatchStatementType.TRIPLES),
```
Karolina-Bogacka, line 23: why for any decoder patchstatementtype.triples

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
```diff
@@ -0,0 +1,354 @@
...
+        RdfPatchRow.ofOptions(JellyPatchOptions.smallStrict
+          .withStatementType(PatchStatementType.QUADS)
+          .withStreamType(PatchStreamType.FLAT)
+        ),
+      ))
+      val decoder = MockPatchConverterFactory.triplesDecoder(null, None)
+      val e = intercept[RdfProtoDeserializationError] {
+        decoder.ingestFrame(input)
+      }
+      e.getMessage should include("Incoming stream with statement type")
```
Karolina-Bogacka, line 244: eeee, matching error message with strings in this way, didn't we want to make it nicer

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
```diff
@@ -0,0 +1,354 @@
...
+      decoder.ingestFrame(input)
+      decoder.getPatchOpt shouldBe Some(opt)
+    }
+
+    "decode quads" in {
+      val input = PatchTestCases.Quads1.encodedFull(
+        JellyPatchOptions.smallStrict
+          .withStatementType(PatchStatementType.QUADS)
+          .withStreamType(PatchStreamType.FLAT),
+        10_000
```
Karolina-Bogacka, line 318: you're using this specific value of 10_000 many times, don't you want to set it as a val earlier in the tests

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/helpers/PatchTestCases.scala
```diff
@@ -0,0 +1,311 @@
...
+    def encoded(opt: RdfPatchOptions): Seq[RdfPatchRow] = Seq(
+      R.ofOptions(opt),
+      R.ofPrefix(RdfPrefixEntry(0, "https://test.org/test/")),
+      R.ofName(RdfNameEntry(0, "subject")),
+      R.ofHeader(RdfPatchHeader("key", RdfIri(1, 0))),
+      R.ofTransactionStart,
+      R.ofName(RdfNameEntry(0, "predicate")),
+      R.ofPrefix(RdfPrefixEntry(0, "https://test.org/ns2/")),
+      R.ofName(RdfNameEntry(0, "object")),
+      R.ofStatementAdd(RdfQuad(
```
Karolina-Bogacka, line 80: I would move some of this sequence building logic to dedicated helpers, maybe at least the triple and quad construction logic - you could have a helper class keeping track of the statements it generated and either returning one of the existing statements to you (so that you could test deleting it) or generate a new statement by having a new subject, predicate or object. Otherwise the testing sequences may be a bit hard to read/maintain, but I guess if the patterns do not overlap enough you may now want to do it

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoder.scala
Ostrzyciel: No, it will. All implementations must behave the same in this regard. The different impls are only for different statement types.

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/internal/PatchDecoderImpl.scala
Ostrzyciel: I like it this way. Making a pattern like you described would add overhead, we can't have that here. Keep in mind that this code is very different from jelly-cli, having very high cyclomatic test coverage, and this behavior is covered there.

core-patch/src/main/scala/eu/ostrzyciel/jelly/core/patch/internal/PatchDecoderImpl.scala
Ostrzyciel: I guess I could simplify that. I'll see what I can do here.

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
Ostrzyciel: Because it must decode something. Only triples and quads are valid values.

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
Ostrzyciel: That would require a complete rewrite of exception handling in Jelly. You sure?

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
Ostrzyciel: No, it's tests. The value doesn't matter, it just needs to be big.

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/helpers/PatchTestCases.scala
Ostrzyciel: I'm not convinced that helpers would be much more maintainable. There are lots of very, very different test cases here. Writing helpers for that would be an utter headache.

core-patch/src/test/scala/eu/ostrzyciel/jelly/core/patch/PatchDecoderSpec.scala
Karolina-Bogacka: eh, let's make it a suggestion for later

Issue #338, opened 2025-04-15
Imprecise error handling


adanilenka:
An invalid jelly file as input raises an Unknown Error.

For the input file, please refer to the test case from: https://github.com/Jelly-RDF/jelly-protobuf/tree/main/test/rdf/from_jelly/triples_rdf_1_1/neg_008

The test attempted to send a name row that exceeded max table size, i.e., table size was set to 16, but there were 17 consecutive name rows with id = 0.

Ostrzyciel:
Stack trace: 

![Image](https://github.com/user-attachments/assets/e01654d4-c7d1-404f-a080-48cc50f34e26)

The issue is that's not very informative. Better would be to tell the user that the id is out of range.

Ostrzyciel:
Should be resolved in #339. Will be part of the next jelly-cli release.

PR #339, opened 2025-04-15
core: improve error handling in decoder & transcoder

Ostrzyciel:
Instead of throwing random NPEs and arrayoutofbounds, throw something nicer. This covers the some of the most common types of exceptions, should be enough.

This should address the complaints made here:

- https://github.com/Jelly-RDF/cli/pull/81#discussion_r2044753735 (will require also changes in cli, but I will do that in a moment)
- #338

### DIFF
--- a/core/src/main/java/eu/ostrzyciel/jelly/core/internal/NameDecoderImpl.java
+++ b/core/src/main/java/eu/ostrzyciel/jelly/core/internal/NameDecoderImpl.java
@@ -123,8 +123,11 @@ final class NameDecoderImpl<TIri> implements NameDecoder<TIri> {
             nameEntry = nameLookup[lastNameIdReference];
         } catch (ArrayIndexOutOfBoundsException e) {
             throw JellyExceptions.rdfProtoDeserializationError(
-                    "Encountered an invalid name table reference (out of bounds). " +
-                            "Name ID: " + nameId + ", Prefix ID: " + iri.prefixId()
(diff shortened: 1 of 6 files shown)


core/src/main/java/eu/ostrzyciel/jelly/core/internal/NameDecoderImpl.java
```diff
@@ -123,8 +123,11 @@ public TIri decode(RdfIri iri) {
             nameEntry = nameLookup[lastNameIdReference];
         } catch (ArrayIndexOutOfBoundsException e) {
             throw JellyExceptions.rdfProtoDeserializationError(
-                    "Encountered an invalid name table reference (out of bounds). " +
-                            "Name ID: " + nameId + ", Prefix ID: " + iri.prefixId()
+                "Encountered an invalid name table reference (out of bounds). " +
```
nk2IsHere, line 126: btw format and multiline strings can be used to make it look a tad nicer :)
```java
var a = 1;
var b = 2;
var s = """
Yes i'm multiline.
Here's a param (%s)
And another one (%s)
""".formatted(a, b);
```
is a valid core for j17

Ostrzyciel: good thing intellij has a quickfix for that, otherwise I'd be too lazy to do it, thanks :D

nk2IsHere (approved):

PR #341, opened 2025-04-16
Core module java rewrite

nk2IsHere:
This PR adds `core-java` and `rdf-protos-java`  modules intended to replace `core` and `rdf-protos`, written in scala to ones written in pure java. 

This is a part of preparation for `v3.0.0` release. 
Link to repo before breaking changes: [v2.10.2](https://github.com/Jelly-RDF/jelly-jvm/tree/d90f502d31561d0a693fee1e8a899b6d33df214c)

Notable API Changes:
- new package naming scheme starting with `eu.neverblink` is introduced to all new modules
- `JellyConstants`, `JellyOptions` with `UPPERCASE_NAMING` for constants replaced `Constants` and `Options`
- `JellyConverterFactory` and `JellyTranscoderFactory` introduced to centralise core modules context maintaining.
- `ProtoDecoder` chain now depends on `ProtoHandler<TNode>` as a handler of different statement types decoding.
- `ProtoDecoder` now does not specify types of Triples and Quads.
- `ProtoEncoder` now does not return `Iterable<Row>` on ingestion, instead a `Collection<RdfStreamRow> appendableRowBuffer` must be passed through parameters for the purposes of encoded statements collection.


build.sbt
```diff
@@ -83,6 +84,19 @@ lazy val rdfProtos = (project in file("rdf-protos"))
     publishArtifact := false,
   )
 
+// Intermediate project that generates the Scala code from the protobuf files
+lazy val rdfProtosJava = (project in file("rdf-protos-java"))
+  .enablePlugins(ProtobufPlugin)
+  .settings(
+    name := "jelly-javameta",
```
Ostrzyciel, line 91:
```suggestion
    name := "jelly-protos-java",
```

the scalameta here referred to a scala lib for metaprogramming

build.sbt
```diff
@@ -83,6 +84,19 @@ lazy val rdfProtos = (project in file("rdf-protos"))
...
+    libraryDependencies ++= Seq(
+      "com.google.protobuf" % "protobuf-java" % protobufV,
+    ),
+    ProtobufConfig / sourceDirectory := baseDirectory.value / "src" / "main" / "protobuf",
```
Ostrzyciel, line 95: In the scala version we had two dirs:

- protobuf, for Jelly-JVM-specific protos
- protobuf_shared, for the stuff imported from the submodule

The first one was important because it includes protoc options like the package name and that the result should be in multiple files.

Get rid of the scalapb options, keep the rest.

Split the giant 15kLOC `Rdf` class file into multiple with the `java_multiple_files` setting there.

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyConstants.java
```diff
@@ -0,0 +1,31 @@
...
+public class JellyConstants {
+
+    private JellyConstants() {}
+
+    public static final String JELLY_NAME = "Jelly";
+    public static final String JELLY_FILE_EXTENSION = "jelly";
+    public static final String JELLY_CONTENT_TYPE = "application/x-jelly-rdf";
+
+    /**
+     * @deprecated Use {@link #PROTO_VERSION_1_0_X} instead.
```
Ostrzyciel, line 12: Remove all deprecated stuff.

Ostrzyciel: For context: the first release with jelly-jvm rewritten in Java will be 3.0.0

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyException.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public static JellyException rdfProtoTranscodingError(String msg) {
+        return new RdfProtoTranscodingError(msg);
+    }
+
+    public JellyException(String message) {
+        super(message);
+    }
+
+    public static final class RdfProtoDeserializationError extends JellyException {
```
Ostrzyciel, line 21: Maybe move these classes to top level and remove the factory methods.

The factories were there just for Java compat.

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyOptions.java
```diff
@@ -0,0 +1,197 @@
...
+
+    private static void checkTableSize(String name, int size, int supportedSize) {
+        checkTableSize(name, size, supportedSize, 0);
+    }
+
+    private static void checkLogicalStreamType(
+        Rdf.RdfStreamOptions options,
+        Rdf.LogicalStreamType expectedLogicalType
+    ) {
+        var logicalType = options.getLogicalType();
```
Ostrzyciel, line 163: Should we use `final var` in places like this? What is the convention in Java projects?

In C++ the `const` religion is a holy one, I know that. In Scala `val` is simply the default.

Unexpected mutations are horrible, but I'm not sure what is the community approach to this.

core-java/src/main/java/eu/ostrzyciel/jelly/core/NodeEncoder.java
```diff
@@ -0,0 +1,66 @@
...
+     * @return The encoded triple node.
+     */
+    RdfTerm.Triple makeQuotedTriple(RdfTerm.SpoTerm s, RdfTerm.SpoTerm p, RdfTerm.SpoTerm o);
+
+    /**
+     * Encode a default graph node.
+     * @return The encoded default graph node.
+     */
+    static RdfTerm.GraphTerm makeDefaultGraph() {
+        return new RdfTerm.DefaultGraph();
```
Ostrzyciel, line 64: Should be a constant value, singleton, basically. No need to allocate more than 1 instance.

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoDecoder.java
```diff
@@ -0,0 +1,30 @@
...
+        super(
+            datatypeClass,
+            converter,
+            nameDecoder
+        );
+    }
+
+    protected abstract Optional<Rdf.RdfStreamOptions> getStreamOptions();
+    
+    public abstract TOut ingestRowFlat(Rdf.RdfStreamRow row);
```
Ostrzyciel, line 24: Rename ingestRowFlat to ingestRow and keep only this method. The second one was for Scala niceness. We don't need to be nice anymore.

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoDecoder.java
```diff
@@ -0,0 +1,30 @@
...
+
+import eu.ostrzyciel.jelly.core.internal.NameDecoder;
+import eu.ostrzyciel.jelly.core.internal.ProtoDecoderBase;
+import eu.ostrzyciel.jelly.core.proto.v1.Rdf;
+
+import java.util.Optional;
+
+public abstract class ProtoDecoder<TNode, TDatatype, TTriple, TQuad, TOut> extends ProtoDecoderBase<TNode, TDatatype, TTriple, TQuad> {
+    protected ProtoDecoder(
+        Class<TDatatype> datatypeClass,
```
Ostrzyciel, line 11: Gosh, remove this, if only possible.

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoEncoder.java
```diff
@@ -0,0 +1,57 @@
...
+        this.options = params.options;
+        this.enableNamespaceDeclarations = params.enableNamespaceDeclarations;
+        this.appendableRowBuffer = params.appendableRowBuffer;
+    }
+
+    public final Iterable<Rdf.RdfStreamRow> addTripleStatement(TTriple triple) {
+        return addTripleStatement(converter.getTstS(triple), converter.getTstP(triple), converter.getTstO(triple));
+    }
+
+    public abstract Iterable<Rdf.RdfStreamRow> addTripleStatement(TNode subject, TNode predicate, TNode object);
```
Ostrzyciel, line 37: All add* methods should return void. We will always append to the row buffer provided in the constructor.

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoEncoder.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.List;
+
+public abstract class ProtoEncoder<TNode, TTriple, TQuad>
+    extends ProtoEncoderBase<TNode, TTriple, TQuad>
+    implements RowBufferAppender {
+
+    public record Params(
+        Rdf.RdfStreamOptions options,
+        boolean enableNamespaceDeclarations,
+        List<Rdf.RdfStreamRow> appendableRowBuffer
```
Ostrzyciel, line 15: We only ever use the `add(E e)` method. So maybe make this arg into `Collection<RdfStreamRow>`?

core-java/src/main/java/eu/ostrzyciel/jelly/core/RdfTerm.java
```diff
@@ -0,0 +1,461 @@
+package eu.ostrzyciel.jelly.core;
+
+import eu.ostrzyciel.jelly.core.proto.v1.Rdf;
+
+public sealed interface RdfTerm {
```
Ostrzyciel, line 5: This file was here mostly as a hack to reduce redundant encapsulation in the generated proto classes. I'm not sure which parts of it are still relevant, this depends on the codegen'd code.

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/DecoderLookup.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.reflect.Array;
+
+public class DecoderLookup<T> {
+
+    private int lastSetId = -1;
+    private final T[] lookup;
+
+    @SuppressWarnings("unchecked")
+    public DecoderLookup(Class<T> type, int maxEntries) {
+        this.lookup = (T[]) Array.newInstance(type, maxEntries);
```
Ostrzyciel, line 12: Ah, I see, you use the Class<T> here.

Look... maybe just cast the return values in get() to T? Sounds simpler and the JVM shouldn't mind. We can worry about the perf issues if they arise.

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderBase.java
```diff
@@ -0,0 +1,117 @@
...
+    protected abstract int getNameTableSize();
+
+    protected abstract int getPrefixTableSize();
+
+    protected abstract int getDatatypeTableSize();
+
+    protected final TNode convertGraphTerm(RdfTerm.GraphTerm graph) {
+        if (graph == null) {
+            throw new JellyException.RdfProtoDeserializationError("Empty graph term encountered in a GRAPHS stream.");
+        } else if (graph instanceof RdfTerm.Iri iri) {
```
Ostrzyciel, line 40: This is going to be "ass" in terms of performance. We will have to revisit this, but first we should merge *something* and then think about un-assifying this.

For context: the Scala implementation here compiled to a JVM tableswitch instruction (similar to OG C++ switch). As fast as it can be, really.

core-java/src/main/java/eu/ostrzyciel/jelly/core/RdfTerm.java
Ostrzyciel: Alrite, I had a look around the rest of the code and it makes sense how you implemented it here.

This will be a performance issue, because you are allocating extra intermediate objects here. We will have to revisit this later.

Please add this to the "little book of performance angst to tackle later".

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/DecoderLookup.java
Ostrzyciel: Please add this to the "little book of performance angst to tackle later".

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderBase.java
Ostrzyciel: Please add this to the "little book of performance angst to tackle later".

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,292 @@
...
+    extends ProtoDecoder<TNode, TDatatype, TTriple, TQuad, TOut> {
+
+    protected final BiConsumer<String, TNode> namespaceHandler;
+    private Rdf.RdfStreamOptions supportedOptions;
+
+    public ProtoDecoderImpl(
+        Class<TDatatype> datatypeClass,
+        ProtoDecoderConverter<TNode, TDatatype, TTriple, TQuad> converter,
+        NameDecoder<TNode> nameDecoder,
+        BiConsumer<String, TNode> namespaceHandler,
```
Ostrzyciel, line 22: Now, this is the biggest API "what???????" in the core module, IMO.

The API for the decoder is like: RdfStreamRow --> decoder.ingest ---> null or triple or quad

Which is a natural way of doing this in Scala, but leads to funny issues.

- Issue 1: you have to check if what you got out is null, triple, or quad, leading to extra overhead
- Issue 2 (pointed out by you): you need a type union on the method's return type
- Issue 3: at some point we needed to add support for namespace declarations, which are neither triples nor quads, and including them in the same API made zero sense. So, I added an ugly callback handler that you see here.

This is nonsense.

Have a look at the `RDFStream` iface in Jena. I suggest making sth like this, but generic and tuned to the interfaces of our beautiful decoder we see here.

Something like `triple(TNode, TNode, TNode)`

Then, we'd add a param to this constructor like `DecoderHandler<TNode> handler`

and call it `handler.triple(...)` in this class.

Also, here comes the kicker, I'd remove the TTriple and TQuad entirely from the generic ifaces, in encoder and decoder. In the encoder I kind of soft-removed them already (made alternatives). In the decoder it was not possible because we were returning the damn `Null | Triple | Quad` abomination. With a callback iface I described above we can do it. The reason here is that it simplifies the implementation AND allows for libraries that don't have a class for triples/quads (Titanium doesn't!).

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,292 @@
...
+        Rdf.RdfStreamOptions supportedOptions
+    ) {
+        super(datatypeClass, converter, nameDecoder);
+        this.namespaceHandler = namespaceHandler;
+        this.supportedOptions = supportedOptions;
+    }
+
+    @Override
+    protected int getNameTableSize() {
+        return Optional.ofNullable(supportedOptions)
```
Ostrzyciel, line 32: Does Java 17 have a null coalescing operator? If not, I'd prefer an if == null here, really...

build.sbt
```diff
@@ -103,6 +117,29 @@ lazy val core = (project in file("core"))
     commonSettings,
   )
 
+lazy val coreJava = (project in file("core-java"))
```
Ostrzyciel, line 120: Package names put the rewrite's classes in eu.neverblink.*

build.sbt
```diff
@@ -57,6 +57,7 @@ lazy val commonSettings = Seq(
     "-unchecked",
   ),
```
Ostrzyciel, line 58: Merge commit: it must include a note that this a Java rewrite of the code that was previously in Scala. I'd also add a permalink to the repo before this was merged, so that you can navigate back and see the code history somehow.

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,292 @@
...
+            .map(Rdf.RdfStreamOptions::getMaxDatatypeTableSize)
+            .orElse(SMALL_DT_TABLE_SIZE);
+    }
+
+    @Override
+    public Optional<Rdf.RdfStreamOptions> getStreamOptions() {
+        return Optional.ofNullable(supportedOptions);
+    }
+
+    public void setStreamOptions(Rdf.RdfStreamOptions options) {
```
Ostrzyciel, line 56: This was private in the original... and had an if preventing overwriting earlier values.

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,292 @@
...
+    @Override
+    protected int getDatatypeTableSize() {
+        return Optional.ofNullable(supportedOptions)
+            .map(Rdf.RdfStreamOptions::getMaxDatatypeTableSize)
+            .orElse(SMALL_DT_TABLE_SIZE);
+    }
+
+    @Override
+    public Optional<Rdf.RdfStreamOptions> getStreamOptions() {
+        return Optional.ofNullable(supportedOptions);
```
Ostrzyciel, line 53: Just return the nullable thing...... who cares

Ostrzyciel: Also (!!!) you are working here with supported options, not the stream options received. These are two very different things. Please inspect the original code.

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    public void setStreamOptions(Rdf.RdfStreamOptions options) {
+        this.supportedOptions = options;
+    }
+
+    @Override
+    public TOut ingestRowFlat(Rdf.RdfStreamRow row) {
```
Ostrzyciel, line 61: make it return void

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoEncoderImpl.java
```diff
@@ -0,0 +1,115 @@
...
+        NodeEncoder<TNode> nodeEncoder,
+        ProtoEncoderConverter<TNode, TTriple, TQuad> converter,
+        ProtoEncoder.Params params
+    ) {
+        super(nodeEncoder, converter, params);
+        this.rowBuffer = Optional.ofNullable(appendableRowBuffer).orElse(new ArrayList<>());
+    }
+
+    @Override
+    public Iterable<Rdf.RdfStreamRow> addTripleStatement(TNode subject, TNode predicate, TNode object) {
```
Ostrzyciel, line 24: Makes me think – can we make these methods implement the same iface as the one we'd be calling in ProtoDecoder?

See the core-patch module. I have there the PatchHandler iface. It's extended by the encoder, and used by the decoder. Clean, shiny. Can we do the same here?

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: Encoder -> RdfHander, Decoder takes RdfHandler. (look at jelly-patch)

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: des decision: do not make use of Optional, due to pef

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: > the callback thing

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoEncoder.java
nk2IsHere: why does scala return iterable? - new api pass row buffer, iterable is remnant

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyException.java
nk2IsHere: move to top level

build.sbt
nk2IsHere: doned. had to do some copy and append to file magic for it to work :P

build.sbt
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyException.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyConstants.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/JellyOptions.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/NodeEncoder.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoDecoder.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoEncoder.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoDecoder.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/ProtoEncoder.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoEncoderImpl.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: doned

core-java/src/main/java/eu/ostrzyciel/jelly/core/internal/ProtoDecoderImpl.java
nk2IsHere: doned

build.sbt
```diff
@@ -57,7 +57,8 @@ lazy val commonSettings = Seq(
...
   javacOptions ++= Seq(
-    "-Werror",
+    "-source", "17",
+//    "-Werror",
```
Ostrzyciel, line 61: Do we have some warnings? Please silence them with annotations if needed, and re-enable this.

core-java/src/main/java/eu/neverblink/jelly/core/JellyConverterFactory.java
```diff
@@ -0,0 +1,121 @@
...
+ */
+public interface JellyConverterFactory<
+    TNode,
+    TDatatype,
+    TEncoderConverter extends ProtoEncoderConverter<TNode>,
+    TDecoderConverter extends ProtoDecoderConverter<TNode, TDatatype>
+> {
+    /**
+     * To be implemented by subclasses. Returns an instance of ProtoEncoderConverter for the RDF library.
+     * @since 2.7.0
```
Ostrzyciel, line 27: I guess the @since declarations are useless, because these ifaces are not the same as in the original anyway... maybe remove all @since in the rewritten module?

Ostrzyciel (changes requested):

build.sbt
nk2IsHere: impossible. its coming from <3 protobuf <3

core-java/src/main/java/eu/neverblink/jelly/core/JellyConverterFactory.java
nk2IsHere: sure why not

build.sbt
nk2IsHere: annnd i seem to be unable to disable deprecation selectively on it..

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+ */
+public abstract class ProtoEncoder<TNode>
+    extends ProtoEncoderBase<TNode>
+    implements RowBufferAppender, ProtoHandler.AnyProtoHandler<TNode> {
+
+    /**
+     * Parameters passed to the Jelly encoder.
+     * <p>
+     * New fields may be added in the future, but always with a default value and in a sequential order.
+     * However, it is still recommended to use named arguments when creating this object.
```
Ostrzyciel, line 20: This is out of date, I think.

I haven't worked with Java records. Can we add new fields in a backwards-compatible manner? Or would we need explicit constructors for that?

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+     *
+     * @param options options for this stream (required)
+     * @param enableNamespaceDeclarations whether to allow namespace declarations in the stream.
+     *      If true, this will raise the stream version to 2 (Jelly 1.1.0). Otherwise,
+     *      the stream version will be 1 (Jelly 1.0.0).
+     * @param appendableRowBuffer buffer for storing stream rows that should go into a stream frame.
+     *      The encoder will append the rows to this buffer.
+     */
+    public record Params(
```
Ostrzyciel, line 29: Is this implicitly `public static class`? Not sure how it works with records.

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Add an RDF quad statement to the stream.
+     * <p>
+     * If your library does not support quad objects, use `addQuadStatement(s, p, o, g)` instead.
+     *
+     * @param quad quad to add
+     * @throws RdfProtoSerializationError if the library does not support quad objects or
+     *                                    if a serialization error occurs.
+     */
+    public final void addQuadStatement(TNode quad) {
```
Ostrzyciel, line 82: That's nonsense. Nodes can't be quads. This method was supposed to be deleted.

core-java/src/main/java/eu/neverblink/jelly/core/ProtoHandler.java
```diff
@@ -0,0 +1,68 @@
+package eu.neverblink.jelly.core;
+
+import java.util.Collection;
+
+/**
+ * Interface for handling different types of RDF data structures that flow from the decoder.
+ *
+ * @param <TNode> The type of the nodes in the RDF data structure, as bound by library.
+ */
+public interface ProtoHandler<TNode> {
```
Ostrzyciel, line 10: Maybe rename to `RdfHandler`?

Ostrzyciel (changes requested):

core-java/src/main/java/eu/neverblink/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,445 @@
...
+/**
+ * Base class for stateful decoders of protobuf RDF streams.
+ *
+ * @see ProtoDecoder the base (extendable) interface.
+ * @see ProtoDecoderBase for common methods shared by all decoders.
+ *
+ * @param <TNode> the type of the node
+ * @param <TDatatype> the type of the datatype
+ */
+public sealed class ProtoDecoderImpl<TNode, TDatatype> extends ProtoDecoder<TNode, TDatatype> {
```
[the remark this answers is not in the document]
nk2IsHere: yes, they do, since we have to create them in JellyConverterFactory

core-java/src/main/java/eu/neverblink/jelly/core/JellyConverterFactory.java
```diff
@@ -0,0 +1,119 @@
...
+     * To be implemented by subclasses. Returns an instance of ProtoDecoderConverter for the RDF library.
+     */
+    TDecoderConverter decoderConverter();
+
+    /**
+     * Create a new ProtoEncoder.
+     * @param params Parameters for the encoder.
+     * @return encoder
+     */
+    default ProtoEncoder<TNode> encoder(ProtoEncoder.Params params) {
```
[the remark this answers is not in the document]
nk2IsHere: we can make them final, sure.

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
nk2IsHere: we need explicit constructors for that

core-java/src/main/java/eu/neverblink/jelly/core/JellyOptions.java
```diff
@@ -0,0 +1,302 @@
...
+import eu.neverblink.jelly.core.proto.v1.RdfStreamOptions;
+import eu.neverblink.jelly.core.utils.LogicalStreamTypeUtils;
+
+/**
+ * A collection of convenient streaming option presets.
+ * None of the presets specifies the stream type – do that with the .withPhysicalType method.
+ */
+public class JellyOptions {
+
+    private JellyOptions() {}
```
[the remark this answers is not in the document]
nk2IsHere: this is for us to not accidentally instantiate a class whose members are only static. there must be no instances of such class

core-java/src/main/java/eu/neverblink/jelly/core/JellyOptions.java
```diff
@@ -0,0 +1,302 @@
+package eu.neverblink.jelly.core;
+
+import eu.neverblink.jelly.core.proto.v1.LogicalStreamType;
+import eu.neverblink.jelly.core.proto.v1.RdfStreamOptions;
+import eu.neverblink.jelly.core.utils.LogicalStreamTypeUtils;
+
+/**
+ * A collection of convenient streaming option presets.
+ * None of the presets specifies the stream type – do that with the .withPhysicalType method.
```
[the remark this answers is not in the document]
nk2IsHere: nope. forgot to change

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
Ostrzyciel: Then I'm for adding explicit constructors. This will be wordy and awful, I know, but otherwise introducing a new parameter to the constructor would be a breaking change (!!!!) requiring us to bump the version to 4.0.0. That's bad.

core-java/src/main/java/eu/neverblink/jelly/core/RdfHandler.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    default void handleNamespace(String prefix, TNode namespace) {
+        // No-op
+    }
+
+    /**
+     * Extension of the ProtoHandler interface to handle triples.
+     * @param <TNode> The type of the nodes in the RDF data structure, as bound by library.
+     */
+    interface TripleStatementHandler<TNode> extends RdfHandler<TNode> {
```
Ostrzyciel, line 22: Just TripleHandler, please...
QuadHandler
GraphHandler
AnyStatementHandler

Triples + Quads = statements in RDF

core-java/src/main/java/eu/neverblink/jelly/core/RdfHandler.java
```diff
@@ -0,0 +1,85 @@
...
+     * Extension of the ProtoHandler interface to handle Triples and Quads.
+     * @param <TNode> The type of the nodes in the RDF data structure, as bound by library.
+     */
+    interface TripleOrQuadStatementHandler<TNode> extends TripleStatementHandler<TNode>, QuadStatementHandler<TNode> {}
+
+    /**
+     * Extension of the ProtoHandler interface to handle any RDF data structure.
+     * @param <TNode> The type of the nodes in the RDF data structure, as bound by library.
+     */
+    interface AnyStatementHandler<TNode>
```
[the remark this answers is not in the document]
Ostrzyciel: or AnyRdfHandler

core-java/src/main/java/eu/neverblink/jelly/core/RdfProtoSerializationError.java
```diff
@@ -0,0 +1,8 @@
+package eu.neverblink.jelly.core;
+
+public final class RdfProtoSerializationError extends RuntimeException {
```
Ostrzyciel, line 3: missing javadoc

core-java/src/main/java/eu/neverblink/jelly/core/RdfProtoSerializationError.java
```diff
@@ -0,0 +1,8 @@
...
+
+    public RdfProtoSerializationError(String msg) {
```
Ostrzyciel, line 5: missing constructor with cause

core-java/src/main/java/eu/neverblink/jelly/core/RdfProtoTranscodingError.java
```diff
@@ -0,0 +1,8 @@
+package eu.neverblink.jelly.core;
+
+public final class RdfProtoTranscodingError extends RuntimeException {
```
Ostrzyciel, line 3: missing javadoc

core-java/src/main/java/eu/neverblink/jelly/core/RdfProtoTranscodingError.java
```diff
@@ -0,0 +1,8 @@
...
+
+    public RdfProtoTranscodingError(String msg) {
```
Ostrzyciel, line 5: missing constructor with cause

core-java/src/main/java/eu/neverblink/jelly/core/RdfTerm.java
```diff
@@ -0,0 +1,515 @@
...
+
+import eu.neverblink.jelly.core.proto.v1.RdfDefaultGraph;
+import eu.neverblink.jelly.core.proto.v1.RdfGraphEnd;
+import eu.neverblink.jelly.core.proto.v1.RdfGraphStart;
+import eu.neverblink.jelly.core.proto.v1.RdfIri;
+import eu.neverblink.jelly.core.proto.v1.RdfLiteral;
+import eu.neverblink.jelly.core.proto.v1.RdfQuad;
+import eu.neverblink.jelly.core.proto.v1.RdfTriple;
+
+public sealed interface RdfTerm {
```
Ostrzyciel, line 11: Missing javadocs in the entire file. Please add them for all important public stuff.

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
nk2IsHere: We can use factory methods like you did in scala,
or we can use `static Params of(...)` in Params (aka "named" constructors) to achieve same effect

core-java/src/main/java/eu/neverblink/jelly/core/internal/NameDecoderImpl.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Update the name table with a new entry.
+     *
+     * @param nameEntry name row
+     * @throws ArrayIndexOutOfBoundsException if the identifier is out of bounds
+     */
+    @Override
+    public void updateNames(RdfNameEntry nameEntry) {
```
Ostrzyciel, line 73: !!!!!! I'm not seeing the changes from #339 https://github.com/Jelly-RDF/jelly-jvm/commit/d90f502d31561d0a693fee1e8a899b6d33df214c

Please make sure you are including EVERYTHING from main as it is now, as there were some changes since you started implementing this.

core-java/src/main/java/eu/neverblink/jelly/core/internal/ProtoEncoderImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+        final var mainRow = RdfStreamRow.newBuilder()
+            .setNamespace(RdfNamespaceDeclaration.newBuilder().setName(prefix).setValue(iriTerm.toProto()).build())
+            .build();
+
+        rowBuffer.add(mainRow);
+    }
+
+    @Override
+    public void handleNamespace(String prefix, String namespace) {
```
Ostrzyciel, line 95: this is duplicated from the above :D no need to have two, keep the one defined in RdfHandler

core-java/src/main/java/eu/neverblink/jelly/core/utils/TripleEncoder.java
```diff
@@ -0,0 +1,21 @@
...
+public interface TripleEncoder<TNode, TTriple> {
+    /**
+     * Encodes a triple into a representation bound to RDF libraries.
+     *
+     * @param subject the subject of the triple
+     * @param predicate the predicate of the triple
+     * @param object the object of the triple
+     * @return the encoded triple
+     */
+    TTriple encode(TNode subject, TNode predicate, TNode object);
```
Ostrzyciel, line 20: very nice, exactly what I wanted, 👏

core-java/src/main/java/eu/neverblink/jelly/core/internal/ProtoDecoderImpl.java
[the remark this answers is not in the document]
nk2IsHere: https://github.com/Jelly-RDF/jelly-jvm/issues/343

core-java/src/main/java/eu/neverblink/jelly/core/ProtoEncoder.java
Ostrzyciel: Sure, you can go with that.

core-java/src/main/java/eu/neverblink/jelly/core/JellyConverterFactory.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Create a new AnyStatementDecoder.
+     * @param supportedOptions maximum supported options for the decoder. If not provided, this.defaultSupportedOptions
+     *                         will be used. If you want to modify this (e.g., to specify an expected logical stream
+     *                         type), you should always use this.defaultSupportedOptions.withXxx.
+     * @param anyProtoHandler the handler to use for decoding any statements
+     * @return decoder
+     */
+    public final ProtoDecoder<TNode, TDatatype> anyDecoder(
```
Ostrzyciel, line 114: anyStatementDecoder...


because it produces only triples and quads, never graph events

core-java/src/main/java/eu/neverblink/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,439 @@
...
+     * The type of the stream is detected automatically based on the options row,
+     * which must be at the start of the stream. If the options row is not present or the stream changes its type
+     * in the middle, an error is thrown.
+     * <p>
+     * Do not instantiate this class directly. Instead use factory methods in
+     * ConverterFactory implementations.
+     */
+    public static final class AnyStatementDecoder<TNode, TDatatype> extends ProtoDecoderImpl<TNode, TDatatype> {
+
+        private final RdfHandler.AnyRdfHandler<TNode> protoHandler;
```
Ostrzyciel, line 351: AnyStatementHandler

core-java/src/main/java/eu/neverblink/jelly/core/internal/ProtoDecoderImpl.java
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        @Override
+        protected void handleQuad(RdfQuad quad) {
+            delegateDecoder.handleQuad(quad);
+        }
+
+        @Override
+        protected void handleGraphStart(RdfGraphStart graphStart) {
+            delegateDecoder.handleGraphStart(graphStart);
```
Ostrzyciel, line 431: Don't worry, this will not produce a graphStart event on the handler. The delegate is only GraphsAsQuadsDecoder. In Scala this would be nicely typed, but with Java we can't have nice things.

Ostrzyciel (changes requested):

Ostrzyciel (approved):

PR #351, opened 2025-04-29
Update Apache Jena to 5.4.0

Ostrzyciel:
There is one method deprecation that we must mark as ignored to maintain compat with 5.0–5.3.

Supersedes #350


Ostrzyciel:
Tests fail due to Jena dropping support for N-Triples-star, apparently. The new RDF 1.2 syntax is completely different.

RDF-star compatible shorthands can still be parsed by the Turtle parser, which makes sense.

However, there is a difference in semantics, that I'm not yet sure how to tackle.

`<< s1 p1 o1 >> p2 o2` in TTL expands to:

```
_:b1 rdf:reifies <<( s1 p1 o1 )>> .
_:b1 p2 o2 .
```

...which is two triples, instead of one. The `<<( )>>` is a "triple term" and the `NodeFactory` only provides methods to construct this, not the `<< >>` syntactic sugar. So, if I understand correctly, the Turtle parser will now generate two triples instead of one, on our RDF-star test cases. Also, these triple terms are only valid as objects of the `rdf:reifies` property, so we can't just replace all `<< >>` with `<<( )>>`. I have zero idea if the test cases will still work with this.

The way I see it, Jena 5.4.0 can't really process RDF-star data anymore, which is *really* confusing. We can't fully move to RDF 1.2 either, because the spec isn't finalized yet. Maybe we'll have to stay on Jena 5.3 for a while and wait until things stabilize.

Ostrzyciel:
> So, if I understand correctly, the Turtle parser will now generate two triples instead of one, on our RDF-star test cases.

Yup, that's correct. So, the RDF-star test cases fail now.

Having a switch to flip RDF 1.2 behavior on and off would be useful, allowing `<< >>` to be parsed and treated like in 5.3. I'll ask the Jena devs if this is intended to be a stable release or not, and if RDF-star support was ever considered to be "stable", because we are dealing with broken N-Triples-star support here.

Ostrzyciel:
Well, that's confusing...

I think a good approach here would be the following:

- Just live with the fact that Jena 5.3 uses RDF-star, and Jena 5.4 uses draft RDF 1.2, and the two are simply incompatible.
- In the Jena integration we should just follow the migration path used for Turtle-star. 
  - So, instead of generating one triple for `<<s1 p1 o1>> p2 o2`, generate two triples.
- This will require some creative hacking, because I'm not sure if we can even output two triples per one input row. We probably will have to lump this with the Java rewrite and Jelly-JVM 3.0.0.
- There should also be a check if we are on Jena 5.4+, to only then enable this behavior. This way, we will be able to support Jena 5.0–5.3 as well in Jelly-JVM 3.0.0.
- For now, Jelly-JVM 2.10 still works with Jena 5.4, but there are rough edges in RDF-star support and inconsistent behaviors. I think that's fine, given that Jena 5.4 by itself is pretty experimental.
- Jelly-JVM 3.0.0 will officially support Jena 5.x+. Jelly-JVM 2.x will officially support Jena 5.0–5.3.

Ostrzyciel:
> In the Jena integration we should just follow the migration path used for Turtle-star.
>    So, instead of generating one triple for <<s1 p1 o1>> p2 o2, generate two triples.

This will probably also decide on the migration path to RDF 1.2 for the Jelly protocol as a whole. If version of Jelly <= 2, then this behavior should be used. Otherwise...? Well, that depends on the updated protocol spec.

niegrzybkowski:
Superseded by: https://github.com/Jelly-RDF/jelly-jvm/pull/482

Issue #348, opened 2025-04-25
Jelly-JVM 3.0.0 release checklist


Ostrzyciel:
Follow-up of #233 and #181 

- [x] Finish #288 (well, at least some of it)
- [x] Remove the old scala modules, rename the java modules to their names (remove "-java" suffix)
- [x] https://github.com/Jelly-RDF/jelly-jvm/issues/233
- [x] Re-point plugin projects to Java implementations
- [x] Optimize using singleton message instances (e.g., RdfGraphEnd, RdfPatchTransaction*). #366 
- [x] Ensure Ascii format for protobufs is restored. (`ProtoAuxiliarySpec` tests) #364 
- [x] Re-enable the Patch integration tests.
- [x] Re-enable forward compat integration tests.
- [x] Tidy up the mess with `BaseJellyOptions` https://github.com/Jelly-RDF/jelly-jvm/pull/347#discussion_r2061256244
- [x] Ensure the internal Java APIs are marked as `@Internal`
- [x] Ensure Java-only packages are published without the `_3` suffix
- [x] Ensure Java-only packages are published without a dependency on Scala
- [x] Ensure that the Jena plugin still works with Fuseki (content negotiation and all)
- [x] Update documentation: #343 
- [x] Ensure all packages are being published to the `eu.neverblink.jelly` namespace (#233)
- [x] ~~Make sure SNAPSHOT releases are working again.~~ Moved to 3.1.0 #391
- [x] Benchmark the thing and make sure we at least match the performance of the Scala implementation. Look for hot spots in places where they shouldn't be.

nk2IsHere:
+ Ensure Ascii format for protobufs is restored. (`ProtoAuxiliarySpec` tests)

Ostrzyciel:
I moved Jena 5.4 support to #368 (3.1.0), because this release is already complex as it is.

nk2IsHere:
Ensure all packages are being published to the eu.neverblink.jelly namespace - organization is globally set to eu.neverblink.jelly

Ostrzyciel:
Ensure that the Jena plugin still works with Fuseki (content negotiation and all):

Done, checked with Fuseki 5.2.0. Uploading a file in Jelly format works, content negotiation for SPARQL CONSTRUCT also works.

Ostrzyciel:
[...]

PR #357, opened 2025-05-03
Java rewrite: use the new proto compiler

Ostrzyciel:
This hooks up the new protoc plugin with the build infrastructure and adapts the existing Java code to the new proto classes.

The way this new plugin is called is a bit hacky, but I guess we can clean up `build.sbt` later.

There are also a few small imperfections in the code that will need to be resolved before 3.0.0, but that's all listed under #348


core-java/src/main/java/eu/neverblink/jelly/core/JellyConverterFactory.java
```diff
@@ -62,7 +62,7 @@ public final ProtoDecoder<TNode, TDatatype> triplesDecoder(
      * Create a new QuadsDecoder.
      * @param supportedOptions maximum supported options for the decoder. If not provided, this.defaultSupportedOptions
      *                         will be used. If you want to modify this (e.g., to specify an expected logical stream
-     *                         type), you should always use this.defaultSupportedOptions.toBuilder().setXxx.build();.
+     *                         type), you should always use this.defaultSupportedOptions.clone().setXxx.build();.
```
nk2IsHere, line 65: .build() is not requited anymore

Ostrzyciel: oui, good catch, removed that here and elsewhere.

nk2IsHere (approved):

Issue #181, opened 2024-10-15
Scala and the future of Jelly-JVM


Ostrzyciel:
I love Scala. It's such a nice language. But, it also comes with a lot of additional overhead that complicates things in a project like Jelly-JVM.

- I had to work around a lot of performance issues, most of which would not exist in Java. 
- Using Jelly-JVM requires one to use the Scala library (in a specific version), which adds to the bloat and makes things "less nice" for people not using Scala.
- Maintaining compatibility over multiple Scala versions is a pain. See #105 and #150 
- Using Jelly-JVM from Java would not look that great, with all these implicits and whatnots.

I'm wondering if it wouldn't make more sense to someday rewrite the whole thing in Java. We could then add user-facing (optional) APIs in Scala, so that Scala fans would have mostly the same experience as they do today.

Open questions:

- How to handle the namespace changes? Some backwards-incompatible changes will surely be needed.
- How do other libraries do this kind of thing?
- Would wrapping the Java APIs in Scala affect the performance negatively?
- How to design the Java APIs to still look "nice", even without all the powerful features of Scala?

This of course is not a first priority. This may be a good thing to reconsider if we will have more compat-breaking features to lump together in something like a 3.0.0 release.

berezovskyi:
Have not started to use Jelly yet, but keeping an eye on it - great work, hope to get to use it someday soon! I have experience using a Scala lib in a Java project (SHaclEX in Eclipse Lyo), so maybe could offer a few opinions:

> Using Jelly-JVM requires one to use the Scala library (in a specific version), which adds to the bloat and makes things "less nice" for people not using Scala.

I think there is more noise/preference about this one than real issues. If you decide you need a certain library, you bring it and its dependencies. Thousands of people use Spark from Java and don't complain (or complain and still do it because it pays the bills). Lots of software needs Electron and therefore is much more bloated than using your lib will (probably) ever be.

> Maintaining compatibility over multiple Scala versions is a pain.

This one I am actually most curious about. How does it work if I in my Java project depend on two Scala libs (SHaclEX and Jelly Jena, for example), do I have to ensure that they both target the same Scala version (assuming I don't care about bloat but only about correctness)? I see that https://github.com/weso/shaclex/blob/master/build.sbt supports 2.12, 2.13, and 3.2.

> Using Jelly-JVM from Java would not look that great, with all these implicits and whatnots.

> How to design the Java APIs to still look "nice", even without all the powerful features of Scala?

[...]

PR #358, opened 2025-05-04
Add crunchy-protoc-plugin

Ostrzyciel:
Issue: https://github.com/Jelly-RDF/jelly-jvm/issues/181

This merges in the protoc plugin code that used to live here: https://github.com/Jelly-RDF/protoc-java-fast That project will be archived.

This compiler generates highly optimized Java code that is similar to what we had in the Scala version of Jelly-JVM, after extensive metaprogramming and manual modifications.

This PR only covers adding the compiler code, without any tests or hooking it up to the build process --- this will be done in a follow-up PR.


crunchy-protoc-plugin/src/main/scala/eu/neverblink/protoc/java/gen/DescriptorGenerator.scala
```diff
@@ -0,0 +1,180 @@
...
+    val initBlock = CodeBlock.builder
+    initBlock.add(
+      "$T.buildFrom($T.parseFrom($N), new $T[] {})",
+      RuntimeClasses.FileDescriptor,
+      RuntimeClasses.FileDescriptorProto,
+      DescriptorGenerator.getDescriptorBytesFieldName,
+      RuntimeClasses.FileDescriptor,
+    )
+    // any file dependencies
+//    if (info.descriptor.getDependencyCount > 0) {
```
Karolina-Bogacka, line 81: are you sure that you'd like to leave these commented out?

PR #360, opened 2025-05-05
Adjust decoder to use explicit kind constants and calculation

nk2IsHere:


build.sbt
```diff
@@ -190,9 +190,12 @@ lazy val coreJava = (project in file("core-java"))
       val inputDir = (rdfProtosJava / target).value / ("scala-" + scalaVersion.value) /
         "src_managed" / "main" / "compiled_protobuf" /
         "eu" / "neverblink" / "jelly" / "core" / "proto" / "v1"
+
       val outputDir = sourceManaged.value / "main" /
         "eu" / "neverblink" / "jelly" / "core" / "proto" / "v1"
-      val javaFiles = (inputDir * "*.java").get
+
+      val javaFiles = (inputDir * "*.java").get ++ (inputDir * "**" * "*.java").get
```
Ostrzyciel, line 197: Nooo, don't do that. The `patch` proto classes should be included in `core`. Only in `core-patch`. This is important, we don't want to bloat the core module.

core-java/src/main/java/eu/neverblink/jelly/core/internal/proto/HeaderBase.java
```diff
@@ -11,4 +12,16 @@ interface Setters extends HeaderBase {
...
+    byte IRI_FIELD_KIND = 0;
+    byte BNODE_FIELD_KIND = 1;
+    byte LITERAL_FIELD_KIND = 2;
+    byte TRIPLE_TERM_FIELD_KIND = 3;
+
+    Object getValue();
+    byte getValueFieldNumber();
+
+    default int getHeaderValueFieldKind() {
+        return this.getValueFieldNumber() - RdfPatchHeader.H_IRI;
```
Ostrzyciel, line 25: Yeah, here is the problem, you can't have the `RdfPatchHeader` class in `core`.

Ostrzyciel (changes requested):

build.sbt
Ostrzyciel: *should not be included in `core`

PR #365, opened 2025-05-07
Rewrite titanium to java

nk2IsHere:
This PR rewrites Titanium RDF API to java as a separate module.
Integration tests for this modules are also unblocked, and passing.
Due to a change of decoder api to use `RdfHandler`, the API of `TitaniumJellyDecoder` was changed significantly.
The API of `TitaniumJellyEncoder` is also changed to separate getting rows and cleaning rows up into explicit actions.

### DIFF
--- a/titanium-rdf-api-java/src/main/java/eu/neverblink/jelly/convert/titanium/internal/TitaniumEncoderConverter.java
+++ b/titanium-rdf-api-java/src/main/java/eu/neverblink/jelly/convert/titanium/internal/TitaniumEncoderConverter.java
@@ -7,29 +7,26 @@ import eu.neverblink.jelly.core.RdfProtoSerializationError;
 
 /**
  * Converter for translating between Titanium RDF API nodes/terms and Jelly proto objects.
- * <p>
- * Quad class is used here, but is not intended to be used with the encoder.
- * The only reason it's here is to satisfy the type signature of the trait.
  */
 @InternalApi
-public final class TitaniumEncoderConverter implements ProtoEncoderConverter<TitaniumNode> {
+public final class TitaniumEncoderConverter implements ProtoEncoderConverter<Object> {
 
     @Override
-    public void nodeToProto(NodeEncoder<TitaniumNode> encoder, TitaniumNode titaniumNode) {
+    public void nodeToProto(NodeEncoder<Object> encoder, Object titaniumNode) {
         try {
-            switch (titaniumNode.type()) {
-                case IRI -> encoder.makeIri(titaniumNode.asStringValue());
-                case BLANK -> encoder.makeBlankNode(titaniumNode.asStringValue().substring(2)); // remove "_:"
-                case SIMPLE_LITERAL -> encoder.makeSimpleLiteral(titaniumNode.asSimpleLiteral().lex());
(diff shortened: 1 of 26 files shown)


titanium-rdf-api-java/src/main/java/eu/neverblink/jelly/convert/titanium/internal/TitaniumNode.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    default DtLiteral asDtLiteral() {
+        return (DtLiteral) this;
+    }
+
+    sealed interface TitaniumLiteral extends TitaniumNode {}
+
+    // String used to represent IRIs and blank nodes
+    record StringNode(String value) implements TitaniumNode {
```
Ostrzyciel, line 46: I'll be grumpy about this one. We should not encapsulate the String in anything else, if not needed. I know that in Java we can't do algebraic datatypes, but then please just use Object and instance of checks.

Ostrzyciel: The fewer allocations, the better :)

titanium-rdf-api-java/src/main/java/eu/neverblink/jelly/convert/titanium/internal/TitaniumEncoderConverter.java
```diff
@@ -0,0 +1,66 @@
...
+import eu.neverblink.jelly.core.InternalApi;
+import eu.neverblink.jelly.core.NodeEncoder;
+import eu.neverblink.jelly.core.ProtoEncoderConverter;
+import eu.neverblink.jelly.core.RdfProtoSerializationError;
+
+/**
+ * Converter for translating between Titanium RDF API nodes/terms and Jelly proto objects.
+ * <p>
+ * Quad class is used here, but is not intended to be used with the encoder.
+ * The only reason it's here is to satisfy the type signature of the trait.
```
Ostrzyciel, line 12: Quad class is no longer required by this trait. Please remove the second paragraph from this javadoc.

titanium-rdf-api-java/src/test/scala/eu/neveblink/jelly/convert/titanium/IntegrationSpec.scala
```diff
@@ -0,0 +1,93 @@
...
+    .addRows(
+      RdfStreamRow.newInstance()
+        .setOptions(JellyOptions.SMALL_STRICT.clone().setPhysicalType(PhysicalStreamType.QUADS))
+    )
+    .addRows(
+      RdfStreamRow.newInstance()
+        .setName(RdfNameEntry.newInstance().setValue("http://example.org/iri"))
+    )
+    .addRows(
+      RdfStreamRow.newInstance()
```
Ostrzyciel, line 26: Could you reuse the nicer `rdfQuad`-and-so-on API from `core-java` tests? To do this, you have to `.dependsOn(coreJava % "compile->compile;test->test)` on this module in build.sbt.

Ostrzyciel (changes requested):

Ostrzyciel (approved):
amazing work :star_struck:

PR #373, opened 2025-05-11
Optimize the Java ser/des implementations

Ostrzyciel:
A few deep optimizations to the new serializer / deserializer implementations, to hopefully bring it to performance parity with 2.10.x. Turns out that it's a really hard task, if you've previously spent... way too much time optimizing the Scala implementation.

This focuses on the Jena and RDF4J integrations. For the others, I've opened follow-up tasks: #376 #375 

- Serializer: RdfStreamRow instances and their container are reused if possible, using the new ReusableRowBuffer class. This is a light extension of the usual Collection interface, so it should be pretty easy to integrate into other stuff. This reduces heap pressure dramatically. The objects are batch-allocated to make the allocation slightly faster (eh, it's marginal anyway) and to hopefully put the Row instances next to each other in memory, to help with cache locality.
  - This optimization is only used for integrations that fully control the lifetime of RdfStreamFrame and its rows.
- Serializer: RdfTriple and RdfQuad instances are reused if possible using an arena allocator. The idea is similar as with the row buffer.
  - This optimization is only used for integrations that fully control the lifetime of RdfTriple and RdfQuad.
- Serializer: Accessors for serialized size were changed up to make them easier to inline for the JVM. It is now possibly to directly access the cached size, which is useful if we are 100% sure the size was really cached before (see next optimization).
- Serializer: The serialized size of a row is computed right after the row is encoded, and not when it is serialized. This is way faster because after encoding the row is still likely in L1 cache, so calculating its size uses up no memory bandwidth. Afterwards, we just read the cached row size, which is trivial in terms of mem bandwidth in comparison (rows should be next to each other in memory => multiple rows fit in one cache line => *zoomies*).
- Deserializer: RdfStreamRow instances and their container are reused if possible, using the same ReusableRowBuffer thing.
- Deserializer: Parsing code for the contents of RdfStreamRow was optimized to make it faster in case there is exactly one contained message (as it should be!). Instead of lookupswitch, we have tableswitch bytecode instruction, and we also skip a lot of funky ifs. This is pretty similar to how it worked in the Scala implementation.
  - I tried the same optimization for all messages, but then things got slower. It turns out that HEBI really were onto something with their fall-through optimization – it works well for example with RdfIri.

I've also did some other very tiny opts that might help with inlining, such as making things final.

I don't have full perf results yet, but for sure it's faster than when I started optimizing it. Serializer should be faster than 2.10. Deserializer – not sure, it seems to depend on the dataset.


nk2IsHere (approved):
bellissimo mamma mia!

Issue #233, opened 2024-12-12
Package renaming in Jelly-JVM 3.0.0


Ostrzyciel:
Up for discussion: in Jelly-JVM 3.0.0 we could consider renaming some packages/modules to better align with what these modules actually do:

- `jelly-stream` -> `jelly-pekko-streams`
- `jelly-grpc` -> `jelly-pekko-grpc`

The "stream" name part is a bit confusing, because it isn't "a module adding support for streams", but rather just an integration for Apache Pekko Streams. Another thing is that in the future we may have other modules with similar capabilities, so it would be very confusing if we had a module named `jelly-stream`. Same goes for gRPC.

To investigate: can we communicate this renaming somehow to Maven?

PR #385, opened 2025-05-16
Remove all Jelly 2.x scala modules and replace them with java modules

nk2IsHere:
This PR also fixes https://github.com/Jelly-RDF/jelly-jvm/issues/233 for jelly-stream renaming (grpc is removed)


build.sbt
```diff
@@ -24,18 +24,9 @@ lazy val jenaV = "5.3.0"
 lazy val rdf4jV = "5.1.3"
 lazy val titaniumApiV = "1.0.0"
 lazy val titaniumNqV = "1.0.2"
-// !! When updating ScalaPB also change the version of the plugin in plugins.sbt
-lazy val scalapbV = "0.11.17"
```
Ostrzyciel, line 28: Please look at /project. You need to remove the plugin, update build.sbt, and remove the metaprogramming stuff.

build.sbt
```diff
@@ -3,7 +3,7 @@
 lazy val scala2MetaVersion = "2.13.15"
```
Ostrzyciel, line 3: This is not needed anymore.

stream/src/main/scala/eu/neverblink/jelly/stream/DecoderFlow.scala
Ostrzyciel, line 1: Please move the package to eu.neverblink.jelly.pekko.stream

Unless you have a better idea.

Ostrzyciel (changes requested):

stream/src/main/scala/eu/neverblink/jelly/stream/DecoderFlow.scala
nk2IsHere: ...jelly.stream was looking nice too tho

stream/src/main/scala/eu/neverblink/jelly/pekko/stream/DecoderFlow.scala
```diff
@@ -0,0 +1,9 @@
+package eu.neverblink.jelly.pekko.stream
```
Ostrzyciel, line 1: sorry, one last thing – pls also rename the directory to `pekko-stream` to make it more consistent.

Also, in that case, should the module be named `jelly-pekko-stream`, without the `s` at the end? It would also be consistent with package and module naming in Pekko itself.

Ostrzyciel (changes requested):

Ostrzyciel (approved):

PR #386, opened 2025-05-16
Extract base options handling to common functionality

nk2IsHere:
And Annotate all internal classes with InternalApi


Ostrzyciel (approved):
Looks good!

PR #398, opened 2025-05-24
Jena: Fix issues with unregistered format variants

Ostrzyciel:
Because we were setting the name of the variant to the stream name from the options, if you tried to pass options with a custom stream name, you would get a Jena error saying that this variant is not registered.


nk2IsHere (approved):

jena/src/main/java/eu/neverblink/jelly/convert/jena/riot/JellyFormatVariant.java
```diff
@@ -112,7 +112,8 @@ private JellyFormatVariant(
         boolean isDelimited,
         int frameSize
     ) {
-        super(options.getStreamName());
+        // Constant, because all writers can handle all variants
+        super("Variant");
```
nk2IsHere, line 116: maybe extract to constant?

PR #401, opened 2025-05-24
Add JMH benchmarks and a failed RdfIri optimization

Ostrzyciel:
I tried optimizing RdfIri parsing, by applying the same pattern as we use in RdfStreamRow parsing (no fallthrough switches, use tableswitch instead). Unfortunately, it is slower:

```
jmh/Jmh/run -wi 10 -i 10 .*RdfIriParseBench.*

# JMH version: 1.37
# VM version: JDK 23.0.1, OpenJDK 64-Bit Server VM, 23.0.1+11-39
# VM invoker: /home/piotr/.jdks/openjdk-23.0.1/bin/java
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 10 iterations, 10 s each
# Measurement: 10 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                                (idDistribution)  Mode  Cnt      Score     Error  Units
RdfIriParseBench.currentImplementation     16,0.5,128,0.5  avgt   50   8013.556 ± 165.799  ns/op
RdfIriParseBench.currentImplementation  1024,0.5,2048,0.5  avgt   50   9915.374 ± 140.165  ns/op
RdfIriParseBench.currentImplementation  1024,0.1,2048,0.6  avgt   50  10656.719 ±  99.072  ns/op
RdfIriParseBench.currentImplementation  1024,0.6,2048,0.1  avgt   50  10779.323 ± 161.682  ns/op
RdfIriParseBench.fallthroughSwitch         16,0.5,128,0.5  avgt   50   7955.070 ±  84.041  ns/op
RdfIriParseBench.fallthroughSwitch      1024,0.5,2048,0.5  avgt   50   9726.173 ± 107.490  ns/op
RdfIriParseBench.fallthroughSwitch      1024,0.1,2048,0.6  avgt   50  10659.187 ± 131.323  ns/op
RdfIriParseBench.fallthroughSwitch      1024,0.6,2048,0.1  avgt   50  10669.884 ± 125.925  ns/op
RdfIriParseBench.tableSwitch               16,0.5,128,0.5  avgt   50   9199.859 ± 166.052  ns/op
RdfIriParseBench.tableSwitch            1024,0.5,2048,0.5  avgt   50  10249.307 ±  87.230  ns/op
RdfIriParseBench.tableSwitch            1024,0.1,2048,0.6  avgt   50  11563.919 ± 144.036  ns/op
RdfIriParseBench.tableSwitch            1024,0.6,2048,0.1  avgt   50  11208.897 ± 176.949  ns/op
```

`currentImplementation` and `fallthroughSwitch` should be the same. `tableSwitch` is the failed challenger.

So yeah, this doesn't work, but still, I put a lot of effort into setting up the JMH benchmark. I think this will be useful down the line, so it makes sense to commit the JMH setup to the repo.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,7 @@ lazy val titaniumApiV = "1.0.0"
 lazy val titaniumNqV = "1.0.2"
 lazy val protobufV = "4.31.0"
 lazy val javapoetV = "0.7.0"
+lazy val jmhV = "1.37"
 
 lazy val commonSettings = Seq(
   libraryDependencies ++= Seq(
@@ -420,3 +421,17 @@ lazy val examples = (project in file("examples"))
     rdf4j,
     titaniumRdfApi
   )
+
+lazy val jmh = (project in file("jmh"))
+  .enablePlugins(JmhPlugin)
+  .settings(
+    publishArtifact := false,
+    name := "jelly-jmh",
+    description := "JMH benchmarks for Jelly-JVM.",
+    libraryDependencies ++= Seq(
+      "org.openjdk.jmh" % "jmh-core" % jmhV,
+      "org.openjdk.jmh" % "jmh-generator-annprocess" % jmhV,
+    ),
+    commonSettings,
+  )
+  .dependsOn(core)
--- a/build.sbt
+++ b/build.sbt
@@ -432,6 +432,7 @@ lazy val jmh = (project in file("jmh"))
       "org.openjdk.jmh" % "jmh-core" % jmhV,
       "org.openjdk.jmh" % "jmh-generator-annprocess" % jmhV,
     ),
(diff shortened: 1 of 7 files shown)


Karolina-Bogacka (changes requested):
I think adding the benchmarks makes perfect sense, I think it could be nice to add a comment somewhere in the docs or in the code that using `RdfIriParseTableswitch` is not advised from an optimization perspective

Karolina-Bogacka (approved):
Ok then, seems sufficient.

PR #404, opened 2025-05-25
Fix incorrect size calculation of cloned messages

Ostrzyciel:
In a few places in the code we clone messages, change their fields, and then forget to reset the cached size. This leads to weird race conditions such as the one caught here: https://github.com/Jelly-RDF/cli/pull/126

We could either be very careful to always reset the size before modifying the cloned message, or just never copy the cached size in the first place. The second approach is much more resistant to human error, so that's what I'm going with here.

Another argument for this solution is that this type of error causes issues that occur only *sometimes*, so they might slip through the cracks in CI testing. It's better to eliminate the entire class of issues in one go.


Karolina-Bogacka (approved):
I agree with the sentiment that it's better to simplify in this case.

PR #405, opened 2025-05-25
Message parser: optimize recursion limits

Ostrzyciel:
I used JMH's perfasm to see what are the bottlenecks in `RdfIri.mergeFrom`, which is one of the hot spots in the parser. It turns out that most time is spent on load/store operations, which is something that we've already suspected.

Some of these ops we can't do much about – we have to read the input byte stream, for example. But I've also noticed that incrementing and decrementing the recursion limit, as well as maintaining the `CodedInputStream` take up quite a few load/stores. While I can't do much with `CodedInputStream` without rewriting chunks of the protobuf-java library, we can change the strategy for maintaining the recursion limit.

Maybe instead of storing it in a field of a helper object, simply pass it through method arguments? Looking at the generated assembly, the code is generally not making a great use of all gazillion registers of modern CPUs, so this value would likely land in some register, eliminating load/stores completely.

So, I got rid of the `LimitedCodedInputStream` class and simply pass through a counter starting at max depth and decrementing with each nested message. This way we can simply test it for being >0, and we don't have to pass around the max depth as well. This again reduces the amount of data that needs to be passed around.

When I tested it with the RdfIri benchmark, I got this:

```
# JMH version: 1.37
# VM version: JDK 23.0.1, OpenJDK 64-Bit Server VM, 23.0.1+11-39
# VM invoker: /home/piotr/.jdks/openjdk-23.0.1/bin/java
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 10 iterations, 10 s each
# Measurement: 15 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: eu.neverblink.jelly.jmh.RdfIriParseBench.currentImplementation
# Parameters: (idDistribution = 16,0.5,128,0.5)

Benchmark                                   (idDistribution)  Mode  Cnt     Score     Error  Units
RdfIriParseBench.currentImplementation        16,0.5,128,0.5  avgt   15  7611.403 ± 162.008  ns/op
RdfIriParseBench.candidateImplementation      16,0.5,128,0.5  avgt   15  6766.457 ± 145.074  ns/op
```

...but when I examined the generated code, it turned out that the JVM simply removed the recursion depth check, because it was provably always false.

So, I did another benchmark, this time with triples nested to random depth:

```
running (fork) org.openjdk.jmh.Main -f5 -wi 10 -i 15 .*RdfTripleRecursiveBench.*
# JMH version: 1.37
# VM version: JDK 23.0.1, OpenJDK 64-Bit Server VM, 23.0.1+11-39
# VM invoker: /home/piotr/.jdks/openjdk-23.0.1/bin/java
# VM options: <none>
# Blackhole mode: compiler (auto-detected, use -Djmh.blackhole.autoDetect=false to disable)
# Warmup: 10 iterations, 10 s each
# Measurement: 15 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 1 thread, will synchronize iterations


Benchmark                                        Mode  Cnt      Score     Error  Units
RdfTripleRecursiveBench.currentImplementation    avgt   75  40747.996 ± 821.271  ns/op
RdfTripleRecursiveBench.candidateImplementation  avgt   75  31447.941 ± 532.946  ns/op
```

Convincing enough for me.

I'm not exactly sure how large will be the real-world impact – I'm guessing around 5%, which is still nice.

To investigate later in this space: further reducing load/stores in the parser and the serializer:

- Could we eliminate some klass checks? These are always a load... need to see if we can help the JVM inline some polymorphic calls with stronger guarantees.
  - See: https://shipilev.net/blog/2015/black-magic-method-dispatch/
- Make sure we are not accessing the input stream more than we need to. There are some checks in the code to make sure the last tag was X and so on... see if we are not repeating something.
- Are there any other things that could be stored on the call stack, instead of the heap?


jmh/src/main/scala/eu/neverblink/jelly/jmh/RdfIriParseBench.scala
```diff
@@ -53,25 +52,25 @@ class RdfIriParseBench:
...
-
-  @Benchmark
-  @OutputTimeUnit(TimeUnit.NANOSECONDS)
-  @BenchmarkMode(Array(Mode.AverageTime))
-  def tableSwitch(input: BenchInput): Unit =
-    val inputStream = input.getInputStream
-    for i <- 0 until input.size do
-      val iri = RdfIriParseTableswitch.newInstance()
-      ProtoMessage.mergeDelimitedFrom(iri, inputStream)
+//  @Benchmark
```
nk2IsHere, line 60: commented out? should we remove it?

Ostrzyciel: I would keep it, this is dirty benchmarking code anyway, so it doesn't hurt.

I'm taking the approach of Jena where they just push JMH benchmarks in semi-working state to a non-published module.

nk2IsHere (approved):

jmh/src/main/scala/eu/neverblink/jelly/jmh/RdfIriParseBench.scala
Ostrzyciel: eh, ok, removed it, whatever

PR #415, opened 2025-06-02
Retain logical type in streaming implementation for custom options

nk2IsHere:


jena/src/test/scala/eu/neverblink/jelly/convert/jena/riot/JellyWriterFactorySpec.scala
```diff
@@ -115,3 +115,29 @@ class JellyWriterFactorySpec extends AnyWordSpec, Matchers, JenaTest:
         IoUtils.autodetectDelimiting(is).isDelimited should be(false)
       }
     }
+
+  "retain `logicalType` if it is set in options in JellyStreamWriterFactory" in {
```
Ostrzyciel, line 119: This should also be tested with triples – for that it's completely separate code.

Ostrzyciel (changes requested):

PR #419, opened 2025-06-05
Use single row buffer in titanium reader

nk2IsHere:


Ostrzyciel (approved):
sleek and simple, nice

PR #445, opened 2025-07-18
Attempt to fix Codecov

Ostrzyciel:
Something is not working right.


Ostrzyciel:
The reason it failed is that 10 months ago, apparently I started setting up codecov but never finished... and there was a stale secret token for codecov here because of that.

:facepalm:

PR #446, opened 2025-07-18
Add a reactive, non-blocking parser for delimited Jelly streams

Ostrzyciel:
This adds `JellyIo.fromByteStreamDelimited`, a method that can parse a stream of `ByteString` objects (emitted by, for example, Pekko's FileIO) into Jelly frames. The flow automatically rechunks the input stream by reading the varint delimiters and splitting/merging ByteStrings as needed.

This should be useful in applications that rely heavily on non-blocking, reactive stream processing. The direct motivation for this feature is NeverBlink's yet unannounced product that we are working on.

The framing is implemented in a custom GraphStage, largely based on Pekko's code ([source](https://github.com/apache/pekko/blob/947ee49293dd57cb488259efac356accfb5c18d3/stream/src/main/scala/org/apache/pekko/stream/scaladsl/Framing.scala#L209)). Because there is a lot that can break here, it has 100% coverage, aside from low-level stream manipulation routines that were copied directly from Pekko.

### DIFF
--- a/examples/src/main/scala/eu/neverblink/jelly/examples/PekkoStreamsWithIo.scala
+++ b/examples/src/main/scala/eu/neverblink/jelly/examples/PekkoStreamsWithIo.scala
@@ -106,7 +106,7 @@ object PekkoStreamsWithIo extends ScalaExample:
     val readFuture = FileIO.fromPath(File("weather2.jelly").toPath)
       // Convert the byte strings to Jelly frames -- this will automatically detect the delimiters
       // and re-chunk the stream if necessary.
-      .via(JellyIo.fromByteStringsDelimited)
+      .via(JellyIo.fromByteStringsDelimited())
       // Decode the Jelly frames to triples.
       .via(DecoderFlow.decodeTriples.asFlatTripleStream)
       .runWith(Sink.seq)


pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/JellyIoOps.scala
```diff
@@ -48,6 +52,50 @@ object JellyIoOps:
...
+     *
+     * @return Pekko Flow
+     */
+    final def fromByteStreamDelimited: Flow[ByteString, RdfStreamFrame, NotUsed] =
+      Flow[ByteString]
+        .via(protobufFraming(Int.MaxValue))
+        .map(byteString => {
+          val byteStrings = byteString.asByteBuffers
+          val codedInputStream = if byteStrings.size > 1 then
+            CodedInputStream.newInstance(byteStrings.map(_.asReadOnlyBuffer()).asJava)
```
Ostrzyciel, line 73: Aren't I converting the ByteStrings to ByteBuffers twice here? Double check it.

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/JellyIoOps.scala
```diff
@@ -48,6 +52,50 @@ object JellyIoOps:
...
+     *
+     * Using this method, you can read Jelly files from a file or socket in a fully reactive manner,
+     * without any blocking operations.
+     *
+     * @return Pekko Flow
+     */
+    final def fromByteStreamDelimited: Flow[ByteString, RdfStreamFrame, NotUsed] =
+      Flow[ByteString]
+        .via(protobufFraming(Int.MaxValue))
+        .map(byteString => {
```
Ostrzyciel, line 70: Consider factoring out the second stage of this flow to a separate method... then we can also have an implementation for parsing non-delimited bytestrings – this would be useful for things like Kafka.

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/JellyIoOps.scala
```diff
@@ -48,6 +52,50 @@ object JellyIoOps:
...
+     *
+     * The implementation is non-blocking and fully reactive.
+     *
+     * The method is made package-private only for use in tests.
+     *
+     * @param maxMessageSize Maximum allowed size for a Protobuf message.
+     *                       If a message exceeds this size, the stage will fail.
+     * @return
+     */
+    private[stream] final def protobufFraming(maxMessageSize: Int): Flow[ByteString, ByteString, NotUsed] =
```
Ostrzyciel, line 94: Consider making it public – would allow people to split the processing across an async boundary, for example. There is nothing "unsafe" about this method.

Ostrzyciel (commented):
self-review

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/JellyIoOps.scala
nk2IsHere: yeah, isnt it like.. the same in terms of what it does?
CodedInputStream.newInstance has different behaviour for collections of byte arrays and just byte array?

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/ProtobufMessageFramingStage.scala
```diff
@@ -0,0 +1,136 @@
...
+
+      /**
+       * try to push downstream, if failed then try to pull upstream
+       */
+      private def tryPushFrame() = {
+        val buffSize = buffer.size
+        if (buffSize >= frameSize) {
+          pushFrame()
+        } else if (buffSize >= 1) {
+          if (varintDecoder(buffer.iterator)) {
```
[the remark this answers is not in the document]
nk2IsHere: (frankly the whole thing is some crazy byte magic, so up to you if you wanna make it nicer for future)

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/JellyIoOps.scala
Ostrzyciel: It actually does. When called with just one ByteBuffer, it applies extra optimizations. I'll add a comment to clarify this, thanks for pointing it out.

pekko-stream/src/main/scala/eu/neverblink/jelly/pekko/stream/impl/ProtobufMessageFramingStage.scala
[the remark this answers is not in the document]
Ostrzyciel: I... would prefer to keep it as-is, especially because it's largely based on Pekko's code. This is pretty delicate, I don't want to mess up the logic here.

nk2IsHere (approved):

PR #449, opened 2025-07-20
Add two missing examples to tests

Ostrzyciel:
We missed them by accident, should also be tested in CI.

We should not make the same mistake again, as we have codecov now – it will shout at us for doing that. Theoretically we could also use reflection to discover all examples, but I feel like that's overcomplicating things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ target
 
 # Output files of example code
 /*.jelly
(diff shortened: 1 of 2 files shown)


Karolina-Bogacka (approved):
Seems like a fair, minor change and the tests pass, so approve

PR #450, opened 2025-07-20
WIP: try to silence down conformance tests

Ostrzyciel:
experimenting with setting this up...

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -27,6 +27,8 @@ jobs:
           - os: ubuntu-latest
             java: 24
     runs-on: ${{ matrix.os }}
+    env:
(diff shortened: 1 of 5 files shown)


Karolina-Bogacka (changes requested):
Looks good, but have you tested the silencing on different modules?

Ostrzyciel:
> Looks good, but have you tested the silencing on different modules?

What do you mean? It works in CI (as you can see) where we test all modules. Silencing is applied in test code explicitly – see the diff where I injected `withSilencedOutput`, so you control when the silencing happens.

Karolina-Bogacka (approved):
Yes, sorry, that was a misunderstanding on my part

PR #453, opened 2025-07-20
Try to fix cross-module test coverage

Ostrzyciel:


Ostrzyciel:
nope, not working, no idea how to get this to work... will come back to this later

Ostrzyciel:
Still not working. I'm giving up on this, all of these code coverage tools are *ancient* and barely working.

The one good idea I had was to use IntelliJ's code coverage agent: https://github.com/JetBrains/intellij-coverage

It does not claim to support Scala, but apparently it does (?), as IntelliJ IDEA has no problems calculating Scala coverage. There is no plugin for this in sbt, so we would need to hack it... or finally switch to mill, which does have a plugin for this (although I'm not sure if it will work in a non-Kotlin project): https://mill-build.org/api/latest/mill/kotlinlib/kover/KoverModule.html

PR #455, opened 2025-07-22
jena-patch: publish JenaChangesCollector class

Ostrzyciel:
Move `JenaChangesCollector` from tests to the compiled/published code and rewrite it in Java. This will be useful for implementing Jelly-Patch benchmarks here: https://github.com/Jelly-RDF/jvm-benchmarks

It may be later useful for jelly-cli as well (see https://github.com/Jelly-RDF/cli/issues/168 ).

This is tested within integration-tests already. We can add unit tests later as needed – I think this is good enough for now.


Karolina-Bogacka (approved):
Approve, I think the CLI features are a good idea

PR #456, opened 2025-07-22
hotfix: fix factory return type for JenaChangesCollector

Ostrzyciel:
I made the return type of JellyPatchOps.changesCollector into `RDFChanges`... which makes all the special methods related to collecting changes inaccessible.

(sigh)

Sorry, we'll need to make a 3.3.1 hotfix release.


Karolina-Bogacka (approved):
Approved, proceed with the hotfix

PR #458, opened 2025-07-23
PatchEncoder: compute serialized size right after encoding

Ostrzyciel:
This will improve cache locality, as all relevant objects are still in cache. Calculating later the total size of the frame will simply require to walk the rows' cached sizes, not touching their nested structures.

This is the same optimization as the one that we have for Jelly-RDF: https://github.com/Jelly-RDF/jelly-jvm/pull/373


Karolina-Bogacka (approved):
Accept, seems like you changed the encoder to use the precomputed data everywhere you could and the test seems reasonable

PR #459, opened 2025-07-23
ProtoEncoder: make sure sizes are really precomputed

Ostrzyciel:
Related to #458

Added test assertions to make sure that the sizes really are precomputed during encoding. This uncovered three cases where this was not done before, so I fixed that as well.


Ostrzyciel:
> For each row you are just checking whether the cached row size is larger than 0. Are you sure that it's enough to check that vs checking whether the cache size is correct?

Yes, that is enough. The correctness of the size is tested in ser/des tests. If the size is incorrect, the stream will be corrupted.

PR #465, opened 2025-07-25
PatchEncoder: use message allocators to reduce heap pressure

Ostrzyciel:
Same optimization as in #373

Should speed up the serializer by a significant margin.


niegrzybkowski (approved):
Looks ok to merge for me, placate the codecov robotic overlords if necessary.

Might be worth it to break up the constructor if possible, as this will add even more things happening inside.

PR #466, opened 2025-07-25
ProtoMessage: increase OutputStream buffer size

Ostrzyciel:
RDF messages quite commonly contain long strings. If we are writing these to an OutputStream, Protobuf will first try to fit the string in its fixed-size buffer (4kB by default), and, if it fails, will create a wholly new buffer for encoding the string.

This is pretty wasteful. I think the correct solution would be to change this part of the Protobuf implementation to write the string in chunks fitting the buffer, but it may be hard to reconcile with other requirements that this class has.

Instead, we can simply raise the buffer size 2x to accommodate larger strings. Note that if the message is known to be smaller than 8kB, a smaller buffer will be allocated instead.

### DIFF
--- a/core/src/main/java/eu/neverblink/protoc/java/runtime/ProtoMessage.java
+++ b/core/src/main/java/eu/neverblink/protoc/java/runtime/ProtoMessage.java
@@ -8,7 +8,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Collection;
-import java.util.List;
 
 /**
  * Abstract interface implemented by Protocol Message objects.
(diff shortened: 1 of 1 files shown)


core/src/main/java/eu/neverblink/protoc/java/runtime/ProtoMessage.java
```diff
@@ -87,11 +93,11 @@ public final MessageType writeDelimitedTo(CodedOutputStream output) throws IOExc
...
-        int bufferSize = CodedOutputStream.computeUInt32SizeNoTag(size) + size;
-        if (bufferSize > CodedOutputStream.DEFAULT_BUFFER_SIZE) {
-            bufferSize = CodedOutputStream.DEFAULT_BUFFER_SIZE;
-        }
+        final int size = getSerializedSize();
+        final int bufferSize = Integer.min(
+            CodedOutputStream.computeUInt32SizeNoTag(size) + size,
+            MAX_OUTPUT_STREAM_BUFFER_SIZE
+        );
         final var codedOutput = CodedOutputStream.newInstance(output, bufferSize);
```
niegrzybkowski, line 101: I think this will mak

Ostrzyciel:
> I think this will make CodedOutputStream allocate an exact-size buffer for anything <8KB. Can you check?

No, that's quite the opposite.

But I started working on a follow-up anyway in which this will become testable – so, let me close this one and come back soon with a better decoupling and tests :)

PR #467, opened 2025-07-25
Native writers: use a single CodedOutputStream

Ostrzyciel:
Two things are going on here:

## Increase OutputStream buffer size

*(this replaces earlier PR #466, this time with more tests)*

RDF messages quite commonly contain long strings. If we are writing these to an OutputStream, Protobuf will first try to fit the string in its fixed-size buffer (4kB by default), and, if it fails, will create a wholly new buffer for encoding the string.

This is pretty wasteful. I think the correct solution would be to change this part of the Protobuf implementation to write the string in chunks fitting the buffer, but it may be hard to reconcile with other requirements that this class has.

Instead, we can simply raise the buffer size 2x to accommodate larger strings. Note that if the message is known to be smaller than 8kB, a smaller buffer will be allocated instead.

## Reuse a single CodedOutputStream in writers

If we are writing a stream of many frames, we are likely going to need the large buffers at some point anyway. In the current setup, for each frame we are allocating a new CodedOutputStream, along with its 4kB buffer (now: will be 8kB). But, it turns out we can simply reuse the CodedOutputStream to write multiple messages, and nothing wrong happens.

CodedOutputStream does use an int32 counter to keep track of bytes written, but as far as I understand the code, it's fine if it overflows, because Java handles overflow arithmetics correctly (e.g., `x = Int.MaxValue; y = x + 20; y - x == 20` without problems). To be 100% sure, I wrote a test (`LongIoStreamSpec`) to check that and yeah, it works.

### DIFF
--- a/core/src/test/scala/eu/neverblink/protoc/java/runtime/ProtobufUtilSpec.scala
+++ b/core/src/test/scala/eu/neverblink/protoc/java/runtime/ProtobufUtilSpec.scala
@@ -1,0 +1,61 @@
+package eu.neverblink.protoc.java.runtime
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpec
+
+import java.io.OutputStream
+import scala.collection.mutable.ListBuffer
+
+class ProtobufUtilSpec extends AnyWordSpec, Matchers {
+  final class WriteTrackingOutputStream extends OutputStream {
+    val writtenBuffers = ListBuffer[Int]()
+
+    override def write(b: Array[Byte], off: Int, len: Int): Unit = {
+      // We measure the size of the actual buffer passed to the write method.
+      writtenBuffers += b.length
+    }
+
+    override def write(b: Int): Unit =
+      writtenBuffers += 1
+  }
+
+  "ProtobufUtil" should {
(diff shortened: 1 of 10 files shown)


core/src/test/scala/eu/neverblink/protoc/java/runtime/ProtobufUtilSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+    "create a CodedOutputStream with a reduced buffer size" in {
+      val os = new WriteTrackingOutputStream()
+      val codedOutputStream = ProtobufUtil.createCodedOutputStream(os, 123)
+      codedOutputStream should not be null
+      codedOutputStream.writeInt32(7, 31241)
+      codedOutputStream.flush()
+
+      os.writtenBuffers should have size 1
+      // +1 byte for delimiter
+      os.writtenBuffers.head should be(123 + 1)
```
niegrzybkowski, line 44: This is what I meant last review - allocating a bunch of differently sized buffers is probably not the best performance-wise. But I see the new writers that are not recreating CodedOutputStream for each message, so it's all good.

niegrzybkowski (approved):
Looks good!

PR #468, opened 2025-07-25
PatchEncoder: use MessageCollection for row buffering

Ostrzyciel:
First step towards using reusable row buffers – use the MessageCollection class to avoid copying the buffer to RdfPatchFrame. This also allows us to use a reusable frame in the Jena writer, which should also speed things up a little bit.

As a side-quest I double-checked if non-delimited Jelly data is correctly written by native writers and yeah, it is. I also removed a redundant cached size reset, as in non-delimited serialization the cached size is not used for anything.


Ostrzyciel:
> Looks good to me.
> Codecov is complaining about RdfPatchWriterJelly, it could use some tests after it's more stabilized.

Codecov is complaining because it doesnt see cross-module test coverage. I cant figure out a way to fix it... maybe we should just switch to mill instead of the damn sbt.

PR #469, opened 2025-07-26
Implement class-based dispatch for term handling

Ostrzyciel:
RDF terms are polymorphic – each may be an IRI, blank node, a literal, or a triple term. To decide which logic route should we execute for a given term, we've used a separate `byte` field that stored the number of the proto message that was currently set. The actual value is stored in an `Object` field and casted to the appropriate subclass, based on the set field number.

However, this introduces extra overhead in maintaining the field number:

- It takes up one extra slot in objects. For example, on modern JVMs with default settings, this increases the size of `RdfQuad` from 32 to 40 bytes. This leads to less efficient cache usage.
- It forces us to maintain additional state in `ProtoEncoder`, to keep track of the currently encoded term, so that we can update it correctly. This adds computational overhead (extra load/store ops) and also increases the size of ProtoEncoder. Because this class will *always* be in L1, we want it to take up at most one cache line. With these fields, encoders take up between 72 and 80 bytes, which is more than one cache line (64B).
- It doesn't actually save us from casting. The field number lets us quickly decide which polymorphic implementation we should use, but type checks have to be performed afterwards regardless.

I tried fixing this in a few ways – here is my third take. Previous approaches can be found in commit history.

This is a similar solution to how it was implemented in Scala (Jelly-JVM 2.x). We have one field of type `Object` that stores the term's value, and we do dispatch using `instanceof` checks. A few things to note:

- No, we can't have anything more specific than `Object`. One of the possible values for terms is `String`, which is outside of our control. In Scala we could use union types, this is really not an option in Java.
- In Scala we used virtual method dispatch, but that's complicated to implement in Java... and instanceof checks are also fast.
- instanceof checks were optimized to check against the final subclasses of proto message classes – so, we do `term instanceof RdfIri.Mutable` instead of `term instanceof RdfIri`.
  - This doesn't sound like it should make a difference, because the JVM should be able to figure out that RdfIri has only one concrete subclass. But, surprisingly, it does. The optimized version generates three x86 instructions (load klass, compare to constant, jump), while the other generates 9 (!!) instructions including 4 loads. Yuck. At least this is what happens on OpenJDK 23.
- This also happens to simplify encoding/decoding code by a lot, which is nice.

## Performance

I tested deserialization performance on built-in JMH benchmarks (separately: parsing and decoding). Serialization was tested on the Jelly-Patch benchmark ([source code](https://github.com/Jelly-RDF/jvm-benchmarks/tree/6cfcafbabcf3c02a2b8861ac34d992a3cba1aa0e)).

Parameters:

```
jmh/Jmh/run -f5 -wi 10 -i 15 .*RdfStreamFrameDecodeBench.*
jmh/Jmh/run -f5 -wi 10 -i 15 .*RdfTripleRecursiveBench.*
Jmh/run -f5 -wi 15 -i 15 --jvmArgs "-Djelly.patch.input-file=data/bsbm-cdc.jellyp -Djelly.patch.statement-type=TRIPLES -Djelly.patch.max-segments=75000" .*SerBench.*
```

Note that for serialization I used the BSBM dataset which is a *very* pessimistic scenario for Jelly, with a lot of entropy and long strings. So, don't expect huge gains.

**Baseline:**

```
[info] Benchmark                                        Mode  Cnt     Score    Error  Units
[info] RdfStreamFrameDecodeBench.currentImplementation  avgt   75  3006.332 ± 23.781  us/op

[info] Benchmark                                      Mode  Cnt      Score     Error  Units
[info] RdfTripleRecursiveBench.currentImplementation  avgt   75  28614.504 ± 155.930  ns/op

[info] Benchmark       (implementation)  Mode  Cnt           Score          Error  Units
[info] SerBench.jelly             jelly    ss   75  1647186185.000 ± 14971742.690  ns/op
```

**With these changes:**

```
[info] Benchmark                                        Mode  Cnt     Score    Error  Units
[info] RdfStreamFrameDecodeBench.currentImplementation  avgt   75  3064.291 ± 32.155  us/op

[info] Benchmark                                      Mode  Cnt      Score     Error  Units
[info] RdfTripleRecursiveBench.currentImplementation  avgt   75  28594.090 ± 409.779  ns/op

[info] Benchmark       (implementation)  Mode  Cnt           Score          Error  Units
[info] SerBench.jelly             jelly    ss   75  1603908657.720 ± 16467636.911  ns/op
```

Parsing is more or less the same (the difference is within the margin of error). Serialization is decidedly faster.

### DIFF
--- a/core-patch/src/main/java/eu/neverblink/jelly/core/patch/internal/PatchDecoderImpl.java
+++ b/core-patch/src/main/java/eu/neverblink/jelly/core/patch/internal/PatchDecoderImpl.java
@@ -114,7 +114,7 @@ public abstract class PatchDecoderImpl<TNode, TDatatype> extends DecoderBase<TNo
             nsRow.getName(),
             // The value is required for the namespace add operation
             getNameDecoder().decode(valueIri.getPrefixId(), valueIri.getNameId()),
-            convertGraphTermWrapped(nsRow.getGraphFieldNumber() - RdfPatchNamespace.G_IRI, nsRow)
+            convertGraphTermWrapped(nsRow)
         );
     }
 
@@ -124,7 +124,7 @@ public abstract class PatchDecoderImpl<TNode, TDatatype> extends DecoderBase<TNo
             nsRow.getName(),
             // The value is not required for the namespace delete operation, null is fine
             decodeNsIri(valueIri),
-            convertGraphTermWrapped(nsRow.getGraphFieldNumber() - RdfPatchNamespace.G_IRI, nsRow)
+            convertGraphTermWrapped(nsRow)
         );
     }
 
@@ -154,10 +154,7 @@ public abstract class PatchDecoderImpl<TNode, TDatatype> extends DecoderBase<TNo
 
     private void handleHeader(RdfPatchHeader hRow) {
         // No support for repeated terms in the header
-        patchHandler.header(
-            hRow.getKey(),
-            convertTerm(hRow.getValueFieldNumber() - RdfPatchHeader.H_IRI, hRow.getValue())
-        );
+        patchHandler.header(hRow.getKey(), convertTerm(hRow.getValue()));
     }
 
     private void handlePunctuation() {
@@ -263,7 +260,7 @@ public abstract class PatchDecoderImpl<TNode, TDatatype> extends DecoderBase<TNo
                 convertSubjectTermWrapped(statement),
                 convertPredicateTermWrapped(statement),
                 convertObjectTermWrapped(statement),
-                convertGraphTermWrapped(statement.getGraphFieldNumber() - RdfQuad.G_IRI, statement)
+                convertGraphTermWrapped(statement)
             );
         }
 
@@ -273,7 +270,7 @@ public abstract class PatchDecoderImpl<TNode, TDatatype> extends DecoderBase<TNo
                 convertSubjectTermWrapped(statement),
                 convertPredicateTermWrapped(statement),
                 convertObjectTermWrapped(statement),
-                convertGraphTermWrapped(statement.getGraphFieldNumber() - RdfQuad.G_IRI, statement)
+                convertGraphTermWrapped(statement)
             );
         }
     }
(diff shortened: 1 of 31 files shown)


crunchy-protoc-plugin/src/main/scala/eu/neverblink/protoc/java/gen/OneOfGenerator.scala
```diff
@@ -41,17 +41,20 @@ class OneOfGenerator(val info: OneOfInfo):
       .addModifiers(Modifier.PROTECTED)
     field.initializer("null")
     t.addField(field.build)
-    val numberField = FieldSpec.builder(TypeName.BYTE, info.numberFieldName)
-      .addModifiers(Modifier.PROTECTED)
-      .initializer("$L", 0)
-    t.addField(numberField.build)
+    // Class-based oneofs do not need a field for the number, we match by class
+    if !info.usesClassBasedOneof then
```
Karolina-Bogacka, line 45: Please elaborate a bit more on this, why check for class-based oneofs?

core-patch/src/test/scala/eu/neverblink/jelly/core/patch/helpers/PatchTestCases.scala
```diff
@@ -324,9 +324,9 @@ object PatchTestCases:
       rdfPatchRow(opt),
       rdfPatchRow(rdfPrefixEntry(0, "https://test.org/test/")),
       rdfPatchRow(rdfNameEntry(0, "")),
-      rdfPatchRowAdd(rdfPatchNamespace("test", rdfIri(1, 0)).setGDefaultGraph(rdfDefaultGraph())),
```
Karolina-Bogacka, line 327: Are you sure you don't need to add some new tests for this approach to encoding/decoding? If everything works on instanceof, are there no new possible edge cases?

Ostrzyciel: Nope, there are enough tests as it is. I can add some tests for when people misuse the low-level API and set the values to be garbage, but it will simply throw an exception so :shrug: I don't see that as an important thing to cover.

Karolina-Bogacka (approved):

core-patch/src/main/java/eu/neverblink/jelly/core/patch/internal/PatchDecoderImpl.java
```diff
@@ -114,7 +114,7 @@ private void handleNamespaceAdd(RdfPatchNamespace nsRow) {
             nsRow.getName(),
             // The value is required for the namespace add operation
             getNameDecoder().decode(valueIri.getPrefixId(), valueIri.getNameId()),
-            convertGraphTermWrapped(nsRow.getGraphFieldNumber() - RdfPatchNamespace.G_IRI, nsRow)
+            convertGraphTermWrapped(nsRow)
```
nk2IsHere, line 117: finally :D

PR #472, opened 2025-07-26
Native parsers: use a single CodedInputStream

Ostrzyciel:
Very similar optimization to #467

We allocate only one CodedInputStream for the entire delimited stream.

This allows us to make a more efficient use of the buffer (having multiple messages in the buffer are fine). We also remove the need to allocate a new buffer for each frame. Finally, we can also remove the LimitedInputStream wrapper, as the parser is free to consume the entire input stream.

I've marked the old `readStream` method as deprecated, because this implementation is both simpler and faster.

I did benchmark this (entire parsing process as a whole), and while it does seem to be a tiny bit faster, the difference is smaller than the error of the measurement, so... yeah. In any case, it definitely saves on allocations (this I could verify with a profiler), so that's a win.

### DIFF
--- a/rdf4j/src/main/java/eu/neverblink/jelly/convert/rdf4j/rio/JellyParser.java
+++ b/rdf4j/src/main/java/eu/neverblink/jelly/convert/rdf4j/rio/JellyParser.java
@@ -113,11 +113,7 @@ public final class JellyParser extends AbstractRDFParser {
             if (delimitingResponse.isDelimited()) {
                 // Delimited Jelly file
                 // In this case, we can read multiple frames
-                readStream(
-                    delimitingResponse.newInput(),
-                    inputStream -> ProtoMessage.parseDelimitedFrom(inputStream, getReusableFrame),
-                    frame -> buffer.clear()
-                );
+                readStream(delimitingResponse.newInput(), getReusableFrame, frame -> buffer.clear());
             } else {
                 // Non-delimited Jelly file
                 // In this case, we can only read one frame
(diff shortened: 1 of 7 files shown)


Karolina-Bogacka (approved):
Looks good, and the test seems sufficient

rdf4j/src/main/java/eu/neverblink/jelly/convert/rdf4j/rio/JellyParser.java
```diff
@@ -113,11 +113,7 @@ public void handleTriple(Value subject, Value predicate, Value object) {
             if (delimitingResponse.isDelimited()) {
                 // Delimited Jelly file
                 // In this case, we can read multiple frames
-                readStream(
-                    delimitingResponse.newInput(),
-                    inputStream -> ProtoMessage.parseDelimitedFrom(inputStream, getReusableFrame),
-                    frame -> buffer.clear()
-                );
+                readStream(delimitingResponse.newInput(), getReusableFrame, frame -> buffer.clear());
```
nk2IsHere, line 116: so parseDelimitedFrom instantiates a new CodedIS for each call?

Ostrzyciel: Yes. That's how it was in every protobuf library I've seen in Java/Scala. They simply have no native support for this whole "delimited stream" thing, as this is not a part of the protobuf spec.

PR #487, opened 2025-08-05
Package the Jena plugin without protobuf-java

### LINKED ISSUE #435: Package the Jena plugin without protobuf-java
Jena core JAR already depends on protobuf-java, so protobuf is already available in Jena classpath. We should thus exclude the protobuf library from the plugin assembly.

Of course, this must be tested...

We use protobuf-java to a minimal degree, so as long as the major version is the same, we should be fine.

niegrzybkowski:
Closes #435

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -376,6 +376,13 @@ lazy val jenaPlugin = (project in file("jena-plugin"))
       "org.apache.jena" % "jena-arq" % jenaV % "provided,test",
       "org.apache.jena" % "jena-fuseki-main" % jenaV % "provided,test",
     ),
+    // Excludes protobuf-java from the jar, since it's already provided in jena-core.
+    // I couldn't figure out a cleaner way since it's also a dependency of jelly-core.
+    ThisBuild / assemblyMergeStrategy := {
+      case PathList("com", "google", "protobuf", _*) => MergeStrategy.discard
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -376,6 +376,13 @@ lazy val jenaPlugin = (project in file("jena-plugin"))
       "org.apache.jena" % "jena-arq" % jenaV % "provided,test",
       "org.apache.jena" % "jena-fuseki-main" % jenaV % "provided,test",
     ),
+    // Excludes protobuf-java from the jar, since it's already provided in jena-core.
+    // I couldn't figure out a cleaner way since it's also a dependency of jelly-core.
```
Ostrzyciel, line 380: You can do that by applying this: https://www.scala-sbt.org/1.x/docs/Library-Management.html#Exclude+Transitive+Dependencies

Will that work?

Ostrzyciel (changes requested):

build.sbt
niegrzybkowski: It won't since the assembly already skips packaging Jena and its dependencies (`% "provided"`). The problem is jelly-core, so `dependsOn(core)` - I don't know how to tell the assembly plugin to exclude a transitive dependency from a project dependency.

PR #489, opened 2025-08-05
Add CI for RDF4J

niegrzybkowski:
Draft, need a CI run to check if caches work. 
The plugin doesn't work, at least with the `convert` command - I tested that locally - might need a fix anyway

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -140,15 +140,45 @@ jobs:
       shell: bash
       run: sbt -v +jenaPlugin/assembly +rdf4jPlugin/assembly
 
+    - name: Cache Jena
+      id: cache-jena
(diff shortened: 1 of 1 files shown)


Ostrzyciel:
This may be a registration issue... In Jena we needed to do special ServiceLoader shenanigans for this to work in the plugin.

niegrzybkowski:
I think it's just that the `convert` command gives a Writer instead of OutputStream, not sure how to handle it. If I get CI to pass to that point it's going to be clear why it's failing.

niegrzybkowski:
It's an upstream issue. All binary formats are borked when writing, RDF4J .brf also crashes. For now I disabled the writing jelly test.

Ostrzyciel (approved):

PR #512, opened 2025-08-26
Fix neo4j plugin initialization

Ostrzyciel:
Previously it only worked after a call to `jelly.version()`. This is not great, so I added a small hack to get the plugin initialized earlier. Unfortunately I don't have a better idea because neither Neo4j nor Neosemantics provide APIs that could help with this.


Karolina-Bogacka (approved):
Not the greatest fan of the hack, but it's definitely better than having to explicitely call jelly.version()

PR #548, opened 2025-09-27
Fix several minor RDF4J Rio incompatibilities

Ostrzyciel:
Even more fixes to make https://github.com/eclipse-rdf4j/rdf4j/issues/5449 possible:

- Enable namespace declarations by default, as all RDF4J writers do that.
- In case the RDF handler in the parser is unset, don't crash, just send the data to space.
- Invoke `clear()` after and before parsing to make sure the base machinery can reset e.g., blank node hashes.


Karolina-Bogacka (approved):
The changes make sense

PR #553, opened 2025-09-28
Only perform recursion depth checks on recursive messages

Ostrzyciel:
A truly micro optimization.

We were checking the recursion depth already quite efficiently, but we can do a little bit better with the observation that most messages are, in fact, not recursive. For them it's enough to just increment the depth.

If we remove the check on everything aside from RdfTriple, our security will not be compromised (as proven by tests). During inlining, the compiler should also get rid of this parameter entirely in some code paths.


Karolina-Bogacka (approved):
The tests seem sufficient and the changes make sense

PR #680, opened 2026-07-01
Upgrade to sbt 2

Ostrzyciel:


Ostrzyciel:
Note for future generations: if sbt breaks again afterwards, DO NOT ATTEMPT TO FIX IT, just replace it with mill.

PR #714, opened 2026-08-18
Add SPARQL support to Jena plugin

Ostrzyciel:
The plugin is now not an alias, but a source-less module that we use only to create a fatjar.


jena-sparql/src/test/scala/eu/neverblink/jelly/convert/jena/sparql/JellySparqlFusekiSpec.scala
```diff
@@ -0,0 +1,122 @@
...
+        null,
+        "N-TRIPLES",
+      )
+    // Port 0 – let the OS pick a free port, so that the test doesn't clash with anything.
+    server = FusekiServer.create().port(0).loopback(true).add("/ds", dataset).build().start()
+
+  override def afterAll(): Unit =
+    if server != null then server.stop()
+
+  // Time out rather than wait forever – a test that hangs blocks the whole build.
```
nk2IsHere: language is weird

nk2IsHere (approved):

PR #739, opened 2026-09-20
sparql+rdf: Catch IRI / value construction exceptions

Ostrzyciel:
This is purely cosmetic, to make the exception classes thrown by the Jelly parser more predictable.

Also clarify one comment about language tags, because some LLM was very upset about us not normalizing them (we should NOT do that).


core-sparql/src/main/java/eu/neverblink/jelly/core/sparql/internal/SparqlDecoderImpl.java
```diff
@@ -154,22 +166,32 @@ public void ingestFrame(SparqlResultsFrame frame) {
             );
         }
 
-        // Decode each variable's column into a row-indexed array (reused across frames)
+        // Decode each variable's column into a row-indexed array (reused across frames).
+        // The try covers the whole column, so it costs nothing per value.
```
nk2IsHere: eh, feels unnecessary comment

nk2IsHere (approved):

